PR #156, opened 2026-07-15
fix: start built shell and restore KDE keybindings

tomjod:
# Pull Request

## Description

Fix Caelestia Shell startup and its KDE/keyd shortcuts using the current `dev` installer flow.

On Fedora, the distro-provided Caelestia CLI can resolve a different Quickshell configuration, so `caelestia shell -d` does not start the shell built by `08-build-shell.sh`. Starting the built `shell.qml` by path fixes startup, but the existing shortcuts still use `qs -c caelestia` or the old CLI. Those commands cannot resolve the path-selected instance and exit with status 255. KDE autostart and keyd can also report different display identifiers, so normal IPC display filtering misses the running instance.

The inline path-based IPC command is too long for keyd's 256-character command limit. This change therefore installs short helpers that restore the target user's session environment and call the exact shell path with `ipc --any-display`.

It also prevents the new installer order from recreating duplicate KWin `Meta+1..5` bindings after keyd has claimed them.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Configuration change
- [ ] Documentation update
- [ ] Refactor / cleanup
- [ ] Other

## Impact

- [x] KDE Plasma settings
- [ ] Quickshell widgets
- [x] Installation / deployment scripts
- [ ] Theme or color system
- [x] Keyboard shortcuts

## Testing

- Rebuilt the commits on current `dev` (`d2f69de`).
- Ran `bash -n` on every changed shell script.
- Ran `.github/scripts/test_repo_integrity.py`.
- Ran `.github/scripts/test_hypr_shim.py`.
- Built `src/bin/hyprctl.cpp` with CMake and Qt 6.
- Generated the keyd configuration and validated it with `keyd -c`; the longest command is 168 characters.
- Tested the runner and IPC helper end-to-end as the desktop user against Plasma Wayland.
- Verified launcher/keybind IPC returns success, KWin D-Bus is reachable, and `Super+Enter` launches Foot.

## Screenshots

Not applicable; this fixes installation and command routing.

## Review status

- [x] Ready for review
- [ ] Work in progress

## Additional context

The branch contains two focused commits: one for the `dev` autostart wrapper and one for keyd/IPC routing. No Darkly, Polonium, or unrelated personal configuration changes are included.


0xSolanaceae (changes requested):
please update the commits to our new installation process in the dev branch :)

send me a review request when you're done

tomjod:
@0xSolanaceae I rebuilt the PR as two clean commits on the current dev branch and adapted the fixes to the new C++ installer flow. Autostart now launches the built shell directly, and keyd shortcuts use a short path-based IPC helper with the correct user session environment. Local integration tests and the repository validation tests pass. Ready for another review.

scripts/10-autostart.sh
```diff
@@ -4,40 +4,42 @@
...
+# on the distro's caelestia-cli version or its config-directory resolution.
 echo "  Creating Caelestia Shell autostart entry..."
 cat > "$HOME/.local/bin/caelestia-autostart.sh" << EOF
 #!/bin/bash
 export QML2_IMPORT_PATH="\$HOME/.local/lib/qt6/qml"
+export CAELESTIA_LIB_DIR="\$HOME/.local/lib/caelestia"
 STATE_DIR="\${XDG_STATE_HOME:-\$HOME/.local/state}/caelestia"
 SCHEME_FILE="\$STATE_DIR/scheme.json"
 i=0
 while [ \$i -lt 15 ] && [ ! -s "\$SCHEME_FILE" ]; do
```
0xSolanaceae: This waits for a file that won't be created anymore (`scheme.json`)

src/bin/caelestia-shell-ipc
```diff
@@ -0,0 +1,28 @@
...
+    QUICKSHELL_PATH="$(command -v quickshell)"
+elif command -v qs >/dev/null 2>&1; then
+    QUICKSHELL_PATH="$(command -v qs)"
+else
+    echo "Quickshell is not installed or is not available in PATH." >&2
+    exit 127
+fi
+
+export QML2_IMPORT_PATH="$HOME/.local/lib/qt6/qml${QML2_IMPORT_PATH:+:$QML2_IMPORT_PATH}"
+export CAELESTIA_LIB_DIR="$HOME/.local/lib/caelestia"
```
0xSolanaceae: this is dead code, no longer used

scripts/10-autostart.sh
0xSolanaceae: No longer error checking if caelestia exists, just silently starts

scripts/10-autostart.sh
```diff
@@ -4,40 +4,42 @@
...
+    QUICKSHELL_PATH=$(command -v quickshell)
+elif command -v qs >/dev/null 2>&1; then
+    QUICKSHELL_PATH=$(command -v qs)
+elif [ -x "/usr/bin/quickshell" ]; then
+    QUICKSHELL_PATH="/usr/bin/quickshell"
+elif [ -x "/usr/local/bin/quickshell" ]; then
+    QUICKSHELL_PATH="/usr/local/bin/quickshell"
 else
-    CAELESTIA_PATH="caelestia"
+    QUICKSHELL_PATH="quickshell"
```
0xSolanaceae: Add in an error message here so it doesn't fail silently

0xSolanaceae (changes requested):

scripts/10-autostart.sh
tomjod: Removed the scheme.json wait entirely. The new installation process no longer creates that file, so the generated autostart wrapper now launches the built shell immediately.

src/bin/caelestia-shell-ipc
tomjod: Removed QML2_IMPORT_PATH and CAELESTIA_LIB_DIR from the IPC helper. They remain only in the autostart wrapper, where they are required to load the built shell modules.

scripts/10-autostart.sh
tomjod: Added explicit installer-time validation for both the built shell entrypoint and the Quickshell executable. Because this follows the new direct-shell startup process, the relevant checks are shell.qml plus quickshell/qs; either missing dependency now exits nonzero and is handled by the installer error dialog.

scripts/10-autostart.sh
tomjod: Replaced the fallback command with an explicit stderr error and exit 127, so a missing Quickshell installation can no longer fail silently.

tomjod:
@0xSolanaceae The two commits are now rebased onto the latest `dev`, and all requested changes have been addressed. I replied to and resolved the four review threads. The repository integrity and hypr shim tests pass locally; the new Validate run is currently marked `action_required` with no jobs, awaiting approval from the base repository. Could you please re-review when ready?

PR #253, opened 2026-07-26
fix(previews): request zkde_screencast so window thumbnails render

WinTone01:
## What

Window thumbnails in the dock hover popup and the window switcher never render — both fall back to drawing the app icon.

## Why

The QML side is already correct: `WindowScreencastRequest` asks KWin for a per-window PipeWire feed over `zkde_screencast_unstable_v1`. The request just never becomes active on a stock install.

KWin gates its privileged Wayland interfaces on the **client's desktop file**. It resolves the client's `/proc/<pid>/exe`, searches installed `.desktop` files for one whose `Exec` resolves to that same binary, and reads `X-KDE-Wayland-Interfaces` from it.

Quickshell's packaged `/usr/share/applications/org.quickshell.desktop` has **no `Exec` line at all**, so the shell matches no service and the interface is never advertised to it. Nothing logs an error — the stream simply stays inactive, which is why this looks like "previews are broken" rather than "previews are not permitted".

## Fix

Ship a per-user override declaring the interface — the same one `org.kde.plasmashell.desktop` requests to draw its own taskbar thumbnails — and install it from the build script, running `kbuildsycoca6` so KWin sees it through ksycoca.

## Verified

On `kwin_wayland` 6.x, after install + shell restart:

```
WindowScreencastGlobal: zkde_screencast_unstable_v1 bound successfully.
WindowScreencastRequest: objectSerial arrived = 3415 for uuid "{630ebc08-...}"
nodeId = 235
```

Thumbnails render live in both the dock hover popup and the window switcher.

## Notes

- `NoDisplay=true` is kept, and does not break the lookup — plasmashell sets it too.
- `qs` is a symlink to `/usr/bin/quickshell`; KWin matches the resolved path, so `Exec` points at `quickshell`.
- The file documents itself, including how to revoke it (delete + `kbuildsycoca6`).
- Touches `scripts/08-build-shell.sh`, as does #(game icons PR) — whichever lands second needs a trivial rebase.

### DIFF
--- a/assets/org.quickshell.desktop
+++ b/assets/org.quickshell.desktop
@@ -1,0 +1,28 @@
+[Desktop Entry]
+Version=1.5
+Type=Application
+NoDisplay=true
+
+Name=Quickshell
+Icon=org.quickshell
+
+# Installed to ~/.local/share/applications, overriding the packaged
+# /usr/share/applications/org.quickshell.desktop.
+#
+# KWin only advertises its privileged Wayland interfaces to clients whose
+# desktop file asks for them: it takes the client's executable (/proc/<pid>/exe),
+# looks for an installed .desktop whose Exec resolves to that same binary, and
+# reads X-KDE-Wayland-Interfaces from it. Quickshell's packaged file has no Exec
+# line at all, so the shell matches nothing and zkde_screencast_unstable_v1 is
+# never offered — live window thumbnails stay empty and fall back to the app
+# icon in both the dock hover popup and the window switcher.
+#
+# zkde_screencast_unstable_v1 is what lets the shell ask KWin for a live
+# PipeWire feed of a single window, which is exactly how Plasma's own Task
+# Manager draws its taskbar thumbnails (org.kde.plasmashell.desktop requests
+# the same interface). org_kde_plasma_window_management goes with it for
+# window metadata.
+#
+# Delete this file to revoke that, then run: kbuildsycoca6
+Exec=/usr/bin/quickshell
+X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1,org_kde_plasma_window_management
--- a/scripts/08-build-shell.sh
+++ b/scripts/08-build-shell.sh
@@ -126,17 +126,6 @@ cmake --install build || {
     exit 1
 }
 
-# Live window thumbnails: KWin hands out zkde_screencast_unstable_v1 only to
-# clients whose desktop file requests it, and Quickshell's packaged entry has no
-# Exec line to match on. Override it per-user, then refresh the KService cache
-# KWin reads. Without this the dock hover popup and window switcher fall back to
-# showing the app icon instead of a live preview.
-if [ -f "$BUNDLE_DIR/assets/org.quickshell.desktop" ]; then
-    mkdir -p ~/.local/share/applications
-    cp --remove-destination "$BUNDLE_DIR/assets/org.quickshell.desktop" ~/.local/share/applications/org.quickshell.desktop 2>/dev/null || true
-    kbuildsycoca6 >/dev/null 2>&1 || true
-fi
-
 # Validate critical QML module presence before declaring success.
 CONFIG_MODULE_DIR="$HOME/.local/lib/qt6/qml/Caelestia/Config"
 if [[ ! -f "$CONFIG_MODULE_DIR/qmldir" ]]; then
--- a/scripts/08-build-shell.sh
+++ b/scripts/08-build-shell.sh
@@ -126,6 +126,17 @@ cmake --install build || {
     exit 1
 }
 
+# Live window thumbnails: KWin hands out zkde_screencast_unstable_v1 only to
+# clients whose desktop file requests it, and Quickshell's packaged entry has no
+# Exec line to match on. Override it per-user, then refresh the KService cache
+# KWin reads. Without this the dock hover popup and window switcher fall back to
+# showing the app icon instead of a live preview.
+if [ -f "$BUNDLE_DIR/assets/org.quickshell.desktop" ]; then
+    mkdir -p ~/.local/share/applications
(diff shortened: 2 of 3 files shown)


scripts/08-build-shell.sh
0xSolanaceae, line 1: This is a build script, file deployment is a config concern and is currently handled by 10-autostart.sh. Also in that same file (lines 84–101) already creates a` ~/.local/share/applications/quickshell.desktop with X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1,org_kde_plasma_window_management` and a correctly resolved `Exec= path, and runs kbuildsycoca6 --noincremental`. Adding the same logic to 08-build-shell.sh creates two places that need to stay in sync.

scripts/08-build-shell.sh
```diff
@@ -126,6 +126,17 @@ cmake --install build || {
...
 }
 
+# Live window thumbnails: KWin hands out zkde_screencast_unstable_v1 only to
+# clients whose desktop file requests it, and Quickshell's packaged entry has no
+# Exec line to match on. Override it per-user, then refresh the KService cache
+# KWin reads. Without this the dock hover popup and window switcher fall back to
+# showing the app icon instead of a live preview.
+if [ -f "$BUNDLE_DIR/assets/org.quickshell.desktop" ]; then
+    mkdir -p ~/.local/share/applications
+    cp --remove-destination "$BUNDLE_DIR/assets/org.quickshell.desktop" ~/.local/share/applications/org.quickshell.desktop 2>/dev/null || true
```
0xSolanaceae: We are currently using `quickshell.desktop`, please fix your name to comply with this

assets/org.quickshell.desktop
```diff
@@ -0,0 +1,28 @@
...
+# icon in both the dock hover popup and the window switcher.
+#
+# zkde_screencast_unstable_v1 is what lets the shell ask KWin for a live
+# PipeWire feed of a single window, which is exactly how Plasma's own Task
+# Manager draws its taskbar thumbnails (org.kde.plasmashell.desktop requests
+# the same interface). org_kde_plasma_window_management goes with it for
+# window metadata.
+#
+# Delete this file to revoke that, then run: kbuildsycoca6
+Exec=/usr/bin/quickshell
```
0xSolanaceae, line 27: This is fragile

0xSolanaceae (changes requested):
I would recommend you close this PR and implement fixes in 10-autostart.sh instead. Good concept but wrong place

When you're done, make a request for me to re-review.

WinTone01:
Thanks for looking. I tried exactly that before answering, and it does not work — here is what I measured, in case it saves you the round trip.

**Adding the key to the autostart entry does not grant the interface.**

I removed my override, put the key on `~/.config/autostart/caelestiashell.desktop`, rebuilt ksycoca, and started the shell through `caelestia-autostart.sh` the way autostart does:

```
WindowScreencastGlobal: zkde_screencast_unstable_v1 is not available
  (isInitialized: false, isActive: false)
WindowScreencastGlobal: cannot request stream for "{7bdeb923-…}" - extension is not active
```

**Why:** KWin does not look at *which* desktop file launched the client. It resolves the client's `/proc/<pid>/exe`, then searches installed `.desktop` files for one whose `Exec` resolves to that same binary, and reads `X-KDE-Wayland-Interfaces` from it.

```
client exe                       /usr/bin/quickshell
caelestiashell.desktop  Exec=    /home/…/.local/bin/caelestia-autostart.sh
```

The autostart entry's `Exec` is our wrapper script, not the quickshell binary, so it never matches. The only file whose `Exec` resolves to `/usr/bin/quickshell` is Quickshell's own `org.quickshell.desktop` — and it ships with **no `Exec` line at all**, which is why nothing matches by default and previews are dead on a stock install.

With the override in place, on the same machine:

```
WindowScreencastGlobal: zkde_screencast_unstable_v1 bound successfully.
WindowScreencastRequest: objectSerial arrived = 12339 …
```

**That said, your "wrong place" point still stands for the install step.** Shipping it from `08-build-shell.sh` is arguable — it is not part of building the shell. I am happy to move the install and `kbuildsycoca6` into `10-autostart.sh`, where the other desktop-file work already lives. That addresses the structural objection and keeps the fix working.

Would you prefer that? Other options, if overriding another project's desktop file is the sticking point rather than the location:

1. Upstream it to Quickshell so `org.quickshell.desktop` ships an `Exec` line and the interfaces — correct long-term, but not something this repo can wait on.
2. Point the autostart entry's `Exec` straight at the quickshell binary and move the wrapper's environment into the entry, so the file we already own is the one KWin matches. I have not tested this; the wrapper exists for `stdbuf` line buffering, so it would need care.

Happy to do whichever you prefer — say the word and I will redo it rather than have you carry the patch.

0xSolanaceae:
@WinTone01 mb, thanks for the correction! Yes, please migrate it to 10-autostart.sh and you will be good to merge.

WinTone01:
Migrated — the install now lives in `scripts/10-autostart.sh` alongside the other desktop-file work, and `08-build-shell.sh` is back to just building. The autostart script resolves `BUNDLE_DIR` the same way the build script does, so it works whether the installer exports it or the script is run on its own.

The file itself still has to be an override of `org.quickshell.desktop` rather than a key on the autostart entry, for the reason above — KWin matches on the client's resolved executable and that entry's `Exec` is the wrapper script.

Verified by deleting the installed file and running `scripts/10-autostart.sh` directly:

```
Requesting KWin screencast interface for window previews...
[OK]  Window preview interface requested.

Exec=/usr/bin/quickshell
X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1,org_kde_plasma_window_management
```

and `kreadconfig6` reading the interfaces back through KService afterwards.

The PR is now just the desktop file and the install step — 55 lines across two files. Ready for re-review whenever you have a moment.

PR #254, opened 2026-07-26
feat(dock): show real icons for games with no desktop entry

WinTone01:
## What

Games launched through a launcher show a generic placeholder in the dock instead of their real icon.

Minecraft via Prism is the clearest case: KWin reports the class as `Minecraft* 1.21.11`, there is no matching desktop entry, and no icon theme ships anything under that name — the icon only exists as a pixmap on the window itself. Many Steam titles hit the same path when `steam_icon_<id>` isn't installed.

## How

For a window with **no resolvable desktop entry**, extract the largest `_NET_WM_ICON` from the window, cache it as a PNG under `~/.cache/caelestia/winicons/`, and use that as the icon source.

Resolution order (`entry icon` → `extracted window icon` → `class as themed name`) lives in a new `qs.services` singleton, `WinIcons`, and both the taskbar tile and the hover popup resolve through `WinIcons.sourceFor()`.

Keeping the map out of the Dock is deliberate rather than stylistic: the Dock is instantiated once per bar, so holding the map per-component meant whichever instance synced last won — an instance that never ran the extractor would clobber the populated map, and the hover popup would show a broken icon while the tile right next to it showed the correct one. A singleton removes that failure mode by construction.

---

## Notes for review — please read before merging

Two things here are worth your call rather than mine:

**1. New runtime dependency.** The extractor is a small Python helper and needs `python-xlib` and `Pillow`. It is invoked lazily — only once per window class, and only for windows with no desktop entry — so nothing runs on a normal session, and if the helper or its imports are missing the dock silently falls back to the current behaviour. Still, it is a dependency the project does not have today. If you would rather not take it, this same extraction is doable in C++ in the existing plugin (an XGetWindowProperty on `_NET_WM_ICON`); happy to redo it that way instead, just say the word.

**2. XWayland only.** `_NET_WM_ICON` is an X11 property, so this covers XWayland clients — which is what the affected games are. Native Wayland apps with no desktop entry that set their icon via `xdg-toplevel-icon` are not covered. In practice I have not hit that combination, but it is a real gap.

Also touches `scripts/08-build-shell.sh`, as does the window-previews PR — whichever lands second needs a trivial rebase.


0xSolanaceae (changes requested):
Great code, this has been bugging me for a while so I appreciate you putting this PR in.

However, as you mentioned, this could be done in C++ via `XGetWindowProperty`. I'd lean toward that approach for a few reasons:

- The python-xlib dependency is non-trivial (not always pre-installed, especially on minimal setups)
- The existing plugin already has the right context (it's a C++ Qt Quick extension linked against the same Qt/KF6 libraries)
- No shelling out to a Python process per window class & it would be about the same length: XGetWindowProperty -> QImage -> save PNG -> emit signal

please rewrite in C++ and resubmit!

(also I am not concerned with 2, you can leave that alone, it's fine as is)

WinTone01:
Agreed on all three points — that was the alternative I offered in the description and your reasoning for it is better than mine. The Python helper was the quick way to prove the idea worked; it should not be how it ships.

I will rewrite it as a C++ extractor in the existing plugin: `XGetWindowProperty` on `_NET_WM_ICON` → `QImage` → cached PNG → signal, dropping `scripts/caelestia-winicon.py`, the per-window-class subprocess and the `python-xlib`/`Pillow` dependency entirely. The plugin does not link X11 yet, so that goes in too.

Two things I would rather confirm than guess at, since you know the codebase better:

1. **Where it belongs.** `Caelestia/Services` alongside the other window-facing code seems right, next to `KWinActiveWindowBridge` — but say if you would rather it sat in `Caelestia/Images` with the other image work.
2. **Linking.** Straight `libX11`, or would you prefer `xcb` to match anything else you have planned? X11 is the simpler call here since `_NET_WM_ICON` is an X property and the windows this targets are XWayland, but I will follow your preference.

Marking this as a draft until the rewrite is up, so it does not sit in your review queue in a state you have already rejected. I will ping you when it is ready.

One note for whenever it lands: this only ever helps XWayland clients, since `_NET_WM_ICON` is an X property. That covers the cases it was written for — Minecraft through Prism, Steam titles without an installed `steam_icon_<id>` — but a native Wayland app with no desktop entry stays on the generic icon either way. Same limitation as the Python version, just worth stating plainly in C++ too.

0xSolanaceae:
@WinTone01 `Caelestia/Services` is the right home for it. For linking, libX11 is fine, no need for xcb.

WinTone01:
Rewritten in C++ — `Caelestia/Services`, `libX11`, as you said.

`WindowIcon::extract()` reads `_NET_WM_ICON` with `XGetWindowProperty`, picks the largest image in the payload, and caches it as a PNG. Windows are matched on `WM_CLASS` and then the title, since KWin reports the title as the class for these — Minecraft comes through as `Minecraft* 1.21.11`.

`scripts/caelestia-winicon.py` is gone, along with the subprocess per window class and the python-xlib and Pillow dependencies.

Two things in the decode that are easy to get wrong, in case you look there first: `_NET_WM_ICON` is `XA_CARDINAL`, so each 32-bit pixel arrives in a 64-bit `long` and has to be narrowed per pixel rather than copied as a block. And the payload is bounds-checked per run, so a truncated property stops the loop instead of reading past the end.

Tested with Minecraft running: 256x256 icon extracted, and it shows in the taskbar.

0xSolanaceae (approved):
LGTM! Please throw in new PR's for more features/bugfixes!

PR #255, opened 2026-07-26
feat(ai): Claude Code, ChatGPT, Gemini and OpenRouter providers

WinTone01:
---

## Fixes folded in after live testing

The providers were then exercised against their real APIs, which turned up several things the first revision got wrong:

- **Tool results broke the agent loop on Gemini.** They were handed back as a pseudo-protocol dump (`Tool: / Command executed: / Output: / Error:`). Gemini's OpenAI-compatible endpoint refuses that: the follow-up turn returns no content and `finish_reason: function_call_filter: MALFORMED_FUNCTION_CALL`, so the assistant announces a search and stops dead. Drop only the empty assistant message and it answers but *imitates the framing*, replying with invented tool output. Results now go back as plain prose.
- **Rate limits are waited out**, using the delay the provider states, with a countdown in the status line. A pending retry belongs to the conversation and model it was scheduled for and is dropped if either changes — otherwise it later answers an abandoned prompt on whatever model is now selected, and spends quota seeding further limits. Per-day quotas are not retried at all, since the suggested delay is seconds while the allowance returns tomorrow.
- **Errors say what actually happened**, including the upstream error OpenRouter nests in `metadata.raw` (its own message is only "Provider returned error").
- **Model lists are fetched once per session**, not on every sidebar open — on a free tier those requests alone could exhaust the quota.
- **Closing the sidebar no longer kills an in-flight answer.**

All verified against the live APIs rather than reasoned about: the malformed-function-call refusal and its fix, the retry-delay parsing (`59.141469639s`), the per-day quota (`GenerateRequestsPerDayPerProjectPerModel-FreeTier`, value 20), and OpenRouter's nested 429.

### DIFF
--- a/shell/plugin/src/Caelestia/Config/aiconfig.hpp
+++ b/shell/plugin/src/Caelestia/Config/aiconfig.hpp
@@ -30,6 +30,36 @@ class AiConfig : public ConfigObject {
     CONFIG_PROPERTY(QString, activeProvider, u"ollama"_s)
     CONFIG_PROPERTY(QString, activeOllamaModel, u"llama3"_s)
 
+    // Master switch for the whole sidebar AI assistant (independent of which
+    // providers are enabled). When false, the AI tab is hidden entirely.
+    CONFIG_PROPERTY(bool, enableAiAssistant, true)
+
+    // Claude Code provider: shells out to the `claude` CLI in headless mode, using
+    // the user's existing subscription login (~/.claude) — no API key required.
+    CONFIG_PROPERTY(bool, enableClaudeCode, true)
+    CONFIG_PROPERTY(QString, claudeCodeBin, u"claude"_s)
+    CONFIG_PROPERTY(QString, defaultClaudeCodeModel, u"default"_s)
+    // Effort / thinking level passed to `claude --effort` ("default" = don't pass).
+    CONFIG_PROPERTY(QString, claudeCodeEffort, u"default"_s)
+
+    // Multiple Claude accounts, each backed by its own CLAUDE_CONFIG_DIR.
+    // claudeAccountsJson: JSON array of {"id","name"} (the default ~/.claude login
+    // is always available as an implicit "Default" entry, id ""). activeClaudeAccount
+    // holds the selected account id ("" = default). loginTerminal is the terminal
+    // emulator used for the interactive `claude` login.
+    CONFIG_PROPERTY(QString, claudeAccountsJson, u"[]"_s)
+    CONFIG_PROPERTY(QString, activeClaudeAccount, u""_s)
+    CONFIG_PROPERTY(QString, loginTerminal, u"konsole"_s)
+
+    // Anthropic (Claude) HTTP API provider (pay-per-token, API key).
+    // Disabled by default; enable to expose it in the provider selector.
+    // The API key is stored here in plaintext; the ANTHROPIC_API_KEY environment
+    // variable takes precedence when set (see getApiKey() in AiAssistant.qml).
+    CONFIG_PROPERTY(bool, enableClaude, false)
+    CONFIG_PROPERTY(QString, anthropicApiKey, u""_s)
+    CONFIG_PROPERTY(QString, anthropicUrl, u"https://api.anthropic.com"_s)
+    CONFIG_PROPERTY(QString, defaultClaudeModel, u"claude-sonnet-5"_s)
+
 public:
     explicit AiConfig(QObject* parent = nullptr)
         : ConfigObject(parent) {}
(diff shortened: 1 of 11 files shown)


shell/plugin/src/Caelestia/Config/aiconfig.hpp
```diff
@@ -30,6 +30,61 @@ class AiConfig : public ConfigObject {
     CONFIG_PROPERTY(QString, activeProvider, u"ollama"_s)
     CONFIG_PROPERTY(QString, activeOllamaModel, u"llama3"_s)
 
+    // Master switch for the whole sidebar AI assistant (independent of which
```
0xSolanaceae, line 33: These keys are all stored in plaintext, please utilize a better fallback that doesn't result in keys being left in plaintext

0xSolanaceae (changes requested):
Again, a great contribution, I would just appreciate if this was split into a few smaller PR's & my one comment is addressed.

WinTone01:
Thanks — both points taken.

## Plaintext keys: fixed

Keys no longer touch `shell.json`. They go into the **Secret Service keyring** (KWallet on KDE, gnome-keyring elsewhere) via `secret-tool`, passed on **stdin** so they never show up in the process list either. Lookup order is otherwise unchanged: the provider's environment variable still wins, then the keyring.

The config fields are kept purely as a migration path — a key already sitting in `shell.json` is moved into the keyring once and the field cleared, so anyone who already entered one is not left with it on disk.

Verified end to end with a real Gemini and OpenRouter key in `shell.json`:

```
before   geminiApiKey: 39 chars   openrouterApiKey: 73 chars
after    geminiApiKey: <empty>    openrouterApiKey: <empty>
keyring  caelestia-ai-gemini: 39 bytes   caelestia-ai-openrouter: 73 bytes
```

One thing worth your call: this adds a **`secret-tool` (libsecret) dependency** for key storage. If you would rather not take it, the alternative I would suggest is dropping in-app key entry entirely and supporting only the environment variables — happy to do that instead, it is strictly less code.

## Splitting the PR

Fair, it is too big to review as one thing. What I would propose, smallest and most independent first:

1. **Provider plumbing** — the provider selector, per-provider URL/model config, and runtime model discovery, with Ollama still the only provider. Small, and everything else builds on it.
2. **Claude Code provider** — the `claude` CLI path, multi-account, and the install/update/login section of the settings page. No API keys involved.
3. **HTTP providers** — Claude API, ChatGPT, Gemini, OpenRouter (one shared OpenAI-compatible path), plus keyring storage.
4. **Agent-loop hardening** — the tool-result format that Gemini rejects, rate-limit handling, and error reporting. These are fixes to (3) and only make sense after it.

(1) and (2) are independently useful; (3) is where the bulk of the review effort actually is.

Does that split look right to you? I would rather agree the boundaries before doing the work than hand you four PRs cut the wrong way. If you would prefer different seams — say, keeping Claude Code and the HTTP providers together, or splitting the settings page out on its own — tell me and I will cut it that way instead.

0xSolanaceae:
@WinTone01 Split looks great, go for it. I am ok with adding the libsecret dependency, just make sure you add it to both the fedora & arch installer.

WinTone01:
`libsecret` is added to the shell package group on both installers — `sdata/arch-dist/installDP.sh` and `sdata/fedora-dist/installDP_fedora.sh`. Confirmed it is the package that actually ships `secret-tool` on Arch; both scripts still parse.

Starting on the split next. I will open them in dependency order and link them back here as they go up:

1. Provider plumbing — selector, per-provider config, runtime model discovery, Ollama still the only provider
2. Claude Code provider — the CLI path, multi-account, install/update/login section
3. HTTP providers — Claude API, ChatGPT, Gemini, OpenRouter on one shared OpenAI-compatible path, plus keyring storage and this libsecret change
4. Agent-loop hardening — tool-result format, rate limiting, error reporting

I will leave this PR open until the last one is up so nothing is lost in between, then close it in favour of the four. Shout if you would rather I close it now and work from the stack instead.

WinTone01:
The split is up:

- #261 — provider seam and runtime model discovery (1/3)
- #262 — Claude Code provider (2/3)
- #263 — HTTP providers, keyring storage, settings page, libsecret (3/3)

Each is stacked on the previous one, and each builds and loads on its own.

It came out as three rather than the four I proposed. The planned fourth — tool-result format, rate-limit handling, error reporting — turned out to be **fixes to code introduced in (3)**, not independent work: separating them would mean shipping a version that is knowingly broken on Gemini and fixing it in the next PR. The reasoning is written up in #263, and I am happy to pull them back out if you would rather review them apart.

Closing this in favour of those three, as agreed.

PR #256, opened 2026-07-26
fix(launcher): unstick the window switcher after a fast Alt+Tab

WinTone01:
## What

Pressing Alt+Tab quickly sometimes leaves the window switcher **open and no longer cycling**.

## Why

`Component.onCompleted` focuses the launcher's search field, but re-opening reuses a `Content` that is already built, so that never runs again — and `onLauncherChanged` set the search text without focusing it. The Alt release that commits the switcher is delivered to that field (`Keys.onReleased`), so with focus elsewhere the release is never seen: the switcher stays open and stops cycling.

## Fix

Re-opening focuses the search field the same way the first open does.

## Verified

Confirmed at runtime: `activeFocus == true` after a re-open, where previously nothing set it at all.

## What was dropped from this PR

Two performance changes were in earlier revisions of this branch and have both been removed after testing:

- **Keeping the launcher content built** (and pre-building it a few seconds into the session), to take construction off the open path. It held everything the launcher had loaded — including full-size wallpaper images — resident for the whole session, and the resulting memory pressure made unrelated things crawl: changing a wallpaper took minutes with the shell over a gigabyte resident.
- **Gating each window preview's screencast request** on `PathView.onPath` plus a settle delay, so the switcher would not ask KWin for a feed for every window at once. This broke the previews.

Both were speculative optimisations that I had not verified against the behaviour they changed, unlike the focus fix. Mentioning them in case either idea looks tempting later — neither is as free as it appears.

## Note

Independent of #253, #254, #255 and #257.

### DIFF
--- a/shell/modules/launcher/Content.qml
+++ b/shell/modules/launcher/Content.qml
@@ -252,6 +252,12 @@ Item {
                             search.text = Visibilities.launcherInitialSearch;
                             Visibilities.launcherInitialSearch = "";
                         }
+                        // Re-opening reuses an already-built Content, which does not
+                        // run Component.onCompleted again — without this the search
+                        // field never regains focus, and since the Alt release that
+                        // commits the window switcher is delivered to this field,
+                        // the switcher would stay open and stop cycling.
+                        search.forceActiveFocus();
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


0xSolanaceae (approved):
Like it, thanks for your contribution!

PR #257, opened 2026-07-26
fix(launcher): stop the previous list rendering on top of the new one

WinTone01:
## What

Open the launcher's keybinds list (`>keybinds`), then clear the search text. The keybinds entries stay on screen and the app list paints on top of them — both sets of rows render superimposed.

The same applies to the animations list.

## Why

Each list's `Loader.active` was set from two places at once:

```qml
onStateChanged: {
    if (state === "keybinds") keybindsList.active = true;
    else keybindsList.active = false;
    ...
}

State {
    name: "keybinds"
    PropertyChanges { target: keybindsList; active: true }
}
```

`PropertyChanges` captures the property's value on state entry so it can restore it on exit. Because the imperative handler had already written `true`, that is the value captured — so leaving the state "restores" `active` back to `true` and the list never unloads. All the list Loaders are `anchors.fill: parent` siblings, so the stale one simply keeps drawing underneath the new one.

## Fix

Give each Loader sole ownership of its own `active`, derived from the state it belongs to, and drop both the `PropertyChanges` and the imperative handler:

```qml
Loader {
    id: keybindsList
    active: root.state === "keybinds"
}
```

Binding to `root.state` rather than to the `show*` flags is deliberate: the state change is deferred by the `Behavior on state` cross-fade, so this keeps the existing fade timing instead of unloading the outgoing list a frame early.

The `implicitWidth`/`implicitHeight` `PropertyChanges` are untouched — only the `active` ones are removed.

## Verified

Instrumented the settled loader values across transitions: exactly one list is active in each state (`apps` → only `appList`, `keybinds` → only `keybindsList`), where previously the outgoing one stayed loaded. The reporter also confirmed the overlap is gone in normal use.

Independent of #253, #254, #255 and #256 — #256 also touches the launcher but a different file.

### DIFF
--- a/shell/modules/launcher/ContentList.qml
+++ b/shell/modules/launcher/ContentList.qml
@@ -76,19 +76,6 @@ Item {
         }
     }
 
-    onStateChanged: {
-        if (state === "keybinds") {
-            keybindsList.active = true;
-        } else {
-            keybindsList.active = false;
-        }
-        if (state === "animations") {
-            animationsList.active = true;
-        } else {
-            animationsList.active = false;
-        }
-    }
-
     states: [
         State {
             name: "apps"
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
LGTM!

PR #258, opened 2026-07-26
fix(notifs): clear all without hanging the shell

WinTone01:
## What

With a large notification backlog, **Clear all freezes the session** — hard enough that it reads as a crash rather than slowness.

## Why

`clear()` closes notifications one at a time, and `NotifData.close()` rebuilds the whole list each time:

```qml
if (locks.size === 0 && Notifs.list.includes(this)) {   // O(n)
    Notifs.list = Notifs.list.filter(n => n !== this);  // O(n) + a new list
```

That reassignment re-runs every derived binding — `notClosed`, `popups`, the unread counters in the bar, the per-app filters in both notification docks — each a full pass over the list, plus a full view rebuild. Clearing therefore costs O(n) work *per notification*. At ~2000 stored that is millions of list operations and 2000 view rebuilds, back to back on the UI thread.

## Fix

Detach the list in a single assignment before closing anything, so the views update once and each close is just a dismiss and a destroy.

`close()` no longer requires the notification to still be in the list before dismissing and destroying it — otherwise this path would leak every one of them.

## Verified

On a real 2097-notification backlog: clearing completed in **1 ms**, with the shell responsive throughout and the store correctly emptied. Before the change the same operation hung the session.

### DIFF
--- a/shell/services/NotifData.qml
+++ b/shell/services/NotifData.qml
@@ -209,11 +209,17 @@ QtObject {
 
     function close(): void {
         closed = true;
-        if (locks.size === 0 && Notifs.list.includes(this)) {
+        if (locks.size > 0)
+            return; // a view is still animating this one; unlock() closes it later
+
+        // Removing it from the list is separate from dismissing it, so a caller
+        // that has already detached this notification (Notifs.clear(), which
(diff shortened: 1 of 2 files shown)


0xSolanaceae (approved):
LGTM!

PR #259, opened 2026-07-26
wallpaper: decode at screen size, route videos correctly, and preview them

WinTone01:
## What

Changing the wallpaper stutters visibly, and the shell's memory footprint is much larger than the wallpaper's on-disk size suggests.

## Why

`CachingAnimatedImage` in the desktop background never sets `sourceSize`, so the wallpaper is decoded at its native resolution however large that is. With an 8K wallpaper:

- 7680 × 4320 × 4 bytes ≈ **132 MB of pixels** per image
- paid **twice** while the outgoing and incoming images overlap during the reveal
- plus a full-resolution JPEG decode on the CPU each time, and a texture upload of the result

Every other wallpaper view already avoids this — the launcher's `WallpaperItem` and the Nexus `WallItem` both set `sourceSize`. Only the one drawing the largest image at the highest frequency did not.

## Fix

Ask for the screen size, matching what the other views do. JPEG supports scaling during decode, so this is both substantially quicker and far smaller.

## Verified

Instrumented the decode on a 7680×4320 wallpaper across a 2560×1440 and a 1920×1080 screen:

```
sourceSize = 2560x1440   implicit = 2560x1440   (was 7680x4320)
sourceSize = 1920x1080   implicit = 1920x1080
```

`implicitWidth/Height` is the decoded size, so the image really is being scaled down during decode rather than after: roughly **132 MB of pixels down to 15 MB**, per image, per screen.

### DIFF
--- a/shell/modules/background/Wallpaper.qml
+++ b/shell/modules/background/Wallpaper.qml
@@ -283,6 +283,8 @@ Item {
 
 
             CachingAnimatedImage {
+                id: wallpaperImage
+
                 anchors.fill: parent
                 path: img.imagePath
                 visible: !img.isVideoImage && img.imagePath !== ""
@@ -290,6 +292,19 @@ Item {
                 fillMode: AnimatedImage.PreserveAspectCrop
                 source: img.imagePath || ""
                 playing: true
+
+                // Decode at the size actually being drawn. Without this the full
+                // image is decoded whatever its resolution: an 8K wallpaper costs
+                // ~132MB of pixels and a slow decode, twice over while the old and
+                // new ones overlap during the reveal — a visible hitch on every
+                // change. JPEG can scale during decode, so asking for the screen
+                // size makes it both quicker and far smaller. Every other wallpaper
+                // view already does this; only the one drawing the biggest image
+                // did not.
+                sourceSize: {
+                    const dpr = (QsWindow.window as QsWindow)?.devicePixelRatio ?? 1;
+                    return Qt.size(wallpaperImage.width * dpr, wallpaperImage.height * dpr);
+                }
 
                 onStatusChanged: {
                     if (status === Image.Ready && !img.isVideoImage)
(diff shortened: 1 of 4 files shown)


WinTone01:
Added a second, related fix on the same file.

**Setting a video wallpaper left the screen white.** `Img.update()` chose which loader to hand the source to by reading the `isVideoImage` binding, which can still hold the *previous* source's answer at the moment it runs. The video was therefore handed to the image loader:

```
CachingAnimatedImage at Wallpaper.qml[285]: Error Reading Animated Image File ...mp4
```

which cannot decode it, so nothing was drawn. It now asks `root.isVideo(root.source)` about the source being switched to.

Verified that the image loader no longer receives the video path. Full playback could not be exercised from here because the `caelestia wallpaper -f` CLI does not accept video paths — it has to be set from the picker, which assigns `actualCurrent` directly.

WinTone01:
Third related change on this file set: **video wallpapers now get a real preview.**

The pickers were handed the video file itself, which `Image` cannot decode, so every video sat on a loading spinner that never resolved.

`Wallpapers.getThumbnailPath()` already described where a video's still should live — `<cache>/wallpapers/<sha256>/first_frame.png` — but **nothing anywhere wrote that file, and the function had no callers.** So the intent was there and the other half was missing.

It now produces it: one frame extracted with `ffmpeg` on first use, cached at exactly that path, handed to the pickers in place of the video. Extraction is per source and happens once; while it runs `thumbFor()` returns empty, so the launcher keeps its videocam icon and the Nexus spinner is finally telling the truth instead of spinning forever.

Verified with the cache cleared: opening the picker produced a 640×360 `first_frame.png` and drew it, no errors.

Note this adds an **ffmpeg** dependency for video previews — it degrades to the previous placeholder if ffmpeg is missing, but if you would rather not take it, say so and I will gate it behind a setting.

0xSolanaceae (approved):
Awesome, thanks for your contribution!

PR #260, opened 2026-07-26
fix(gamemode): make it work off Hyprland, and pause video wallpapers (1/2)

WinTone01:
## What

On a non-Hyprland compositor, game mode does nothing — and will not even stay switched on.

## Why

Two problems, both from assuming Hyprland is the compositor:

**1. The toggle could not stay on.** The persisted flag was bound to live Hyprland state:

```qml
property bool enabled: Hypr.options["animations:enabled"] === 0
```

With no Hyprland that evaluates `undefined === 0` — false — so the flag reverted as soon as it was set.

**2. Everything it did was a Hyprland call.** `Hypr.extras.applyOptions(...)` to disable animations, blur, shadows, gaps; `Hypr.extras.message("reload")` to restore. Both are no-ops elsewhere.

## Fix

- The flag is plain persisted state. `onConfigReloaded` already re-applies the options on Hyprland, so nothing needed the binding.
- KWin gets the equivalents of what the Hyprland path disables: window animations (`AnimationDurationFactor`) and the blur effect (`Plugins/blurEnabled`), applied with a `reconfigure`. Both are **read back before being changed and restored afterwards**, so a user who already had blur off does not get it switched on when game mode ends.
- Hyprland behaviour is unchanged — the new path is only taken when `HYPRLAND_INSTANCE_SIGNATURE` is unset.

## Also: video wallpapers pause with game mode

A video wallpaper keeps a decoder and a GPU upload running for as long as it plays, which is exactly the load game mode exists to shed. It is now paused while game mode is on, and whether it was *already* paused is remembered — ending game mode does not start one the user had stopped.

## Verified

On `kwin_wayland`, over a full enable/disable cycle:

| | before | game mode on | after |
|---|---|---|---|
| `blurEnabled` | true | **false** | true |
| `AnimationDurationFactor` | 0.25 | **0** | 0.25 |
| `videoWallpaperPaused` | false | **true** | false |

Note the animation factor returns to `0.25` — the user's own non-default value — rather than being clobbered with the default. `gameMode isEnabled` also reports `true` while on, which it never did before.

### DIFF
--- a/shell/services/GameMode.qml
+++ b/shell/services/GameMode.qml
@@ -12,6 +12,16 @@ Singleton {
 
     property alias enabled: props.enabled
 
+    // Hyprland is not always the compositor this runs under. Everything below
+    // branches on that rather than assuming it.
+    // Quickshell.env returns undefined for an unset variable, not "", so compare
+    // truthiness — !== "" was true everywhere and sent KDE down the Hyprland path.
+    readonly property bool onHyprland: !!Quickshell.env("HYPRLAND_INSTANCE_SIGNATURE")
+
+    // Video wallpapers keep a decoder and a GPU upload running for as long as
+    // they play, which is exactly what game mode is trying to free up.
+    property bool restoreVideoWallpaper: false
+
     function setDynamicConfs(): void {
         Hypr.extras.applyOptions({
             "animations:enabled": 0,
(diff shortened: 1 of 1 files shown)


WinTone01:
Added the missing half of the feature.

`utilities.gameMode.autoEnableRegexes` was editable in Settings but **nothing anywhere evaluated it** — it appears only in the config definition and the page that edits the list. So a user could add `Minecraft* 1.21.11`, launch the game, and nothing would happen.

The rules are now matched against the open windows (same source the dock uses: `KWinActiveWindowBridge.windowList`, falling back to `HyprlandData.windowList`), on class, initial class and title. Each rule is tried as a plain name before being treated as a regex, because entries like `Minecraft* 1.21.11` are window titles and not valid patterns — `new RegExp()` on them either throws or matches the wrong thing.

Only a run that switched itself on switches itself back off, so closing a game never undoes a manual toggle.

0xSolanaceae:
@WinTone01 appreciate all your help here! Will check these out ASAP

0xSolanaceae (changes requested):
This code looks great, but you're including auto-enable rules that don't match the scope of this PR. Please split this PR into two PR's, one for the gamemode fix and one for the auto-enable to make it easier for me to review and I can merge each individually. 

When you're done, make a request for me to re-review.

WinTone01:
Split as asked. This PR is now **only the game mode fix** — one file, 62 insertions.

The auto-enable rules moved to a separate PR stacked on this one, linked in a moment.

Re-verified the fix on its own after the split, over a full cycle:

| | before | game mode on | after |
|---|---|---|---|
| `blurEnabled` | true | **false** | true |
| `videoWallpaperPaused` | false | **true** | false |

Ready for re-review.

PR #286, opened 2026-07-28
fix: don't run gdbus through a shell in GlobalShortcut

### LINKED ISSUE #272: Shell command injection in GlobalShortcut via system() + unescaped component names
### Before you submit

### Diagnostics

Both call sites need to go: replace system() with QProcess, passing
program + arguments as a list (QProcess::setProgram("gdbus"),
setArguments({...})) so values are passed as real argv entries instead
of being re-parsed by a shell. That removes the injection vector
structurally instead of trying to escape quotes.

### Issue details

GlobalShortcut::~GlobalShortcut() and GlobalShortcut::updateShortcut()
both build a `gdbus call ...` command by interpolating
info.componentUniqueName() and info.uniqueName() into a string with
QString::arg(), then run it with system():

    QString cmd = QString("gdbus call ... \"['%1', '%2', ...").arg(...).arg(...);
    system(cmd.toUtf8().constData());

componentUniqueName/uniqueName come from KGlobalAccel::globalShortcutsByKey(),
i.e. shortcut registrations from *other* applications on the system,
not from Caelestia. A single quote in either value breaks out of the
quoted gdbus argument, and system() runs the rest through /bin/sh -c
with the user's privileges. Any app that registers a KGlobalAccel
shortcut with a crafted component/action name can run arbitrary shell
commands whenever Caelestia detects a shortcut conflict (updateShortcut)
or shuts down and restores stolen shortcuts (the destructor).

### Confirmation

WinTone01:
Closes #272.

`~GlobalShortcut()` and `updateShortcut()` built a gdbus command line by interpolating `componentUniqueName()` and `uniqueName()` into a string and ran it with `system()`. Those names come from other applications KGlobalAccel registrations, not from Caelestia, so a single quote in either one closes the quoted argument and `/bin/sh` runs whatever follows with the user privileges. It fires whenever a shortcut conflict is detected, or on shutdown when stolen shortcuts are restored.

Both call sites now go through one helper that runs gdbus via `QProcess` with an argument list, so the values become real argv entries and no shell re-parses them. They are also escaped for GVariant text, because gdbus itself parses that argument as GVariant where single quotes delimit strings.

Behaviour is otherwise unchanged: the call still blocks, now with a 2s timeout instead of none.

Checked the escaping output directly — `evil'; rm -rf ~ #` comes out as `evil\'; rm -rf ~ #` inside the GVariant string rather than terminating it.

### DIFF
--- a/shell/plugin/src/Caelestia/Services/globalshortcut.cpp
+++ b/shell/plugin/src/Caelestia/Services/globalshortcut.cpp
@@ -3,9 +3,49 @@
 #include <KGlobalAccel>
 #include <QKeySequence>
 #include <QDebug>
-#include <cstdlib>
+#include <QProcess>
 #include "../Config/config.hpp"
 #include "../Config/generalconfig.hpp"
+
+namespace {
+
+// gdbus parses its arguments as GVariant text, where single quotes delimit
+// strings. Component and action names come from other applications'
+// KGlobalAccel registrations, so they are untrusted input and must be escaped
+// before being placed inside a quoted GVariant string.
+QString escapeGVariantString(const QString &value)
+{
+    QString escaped = value;
+    escaped.replace('\\', QStringLiteral("\\\\"));
+    escaped.replace('\'', QStringLiteral("\\'"));
+    return escaped;
+}
+
+// Passes the arguments as real argv entries so no shell ever re-parses them.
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
LGTM!

PR #287, opened 2026-07-28
fix: escape window addresses in generated KWin scripts

### LINKED ISSUE #270: KWin script generation doesn't escape address before interpolating into JS string literals
### Before you submit

### Diagnostics

Suggested fix: add a small escaping helper (escape backslash and
double-quote at minimum) and route the `.arg(address)` calls in all
nine methods through it before building the script string.

### Issue details

KWinActiveWindowBridge builds ad-hoc KWin JS scripts in focusWindow(),
closeWindow(), minimizeWindow(), maximizeWindow(), raiseWindow(),
moveWindow(), resizeWindow(), setWindowProperty(), and
setWindowDesktop(). All of them interpolate the `address` parameter
directly into a double-quoted JS string literal via QString::arg(),
with no escaping:

    if (wins[i].internalId && String(wins[i].internalId) === "%1") {

If `address` ever contains a `"` character, the generated script is
malformed: everything after the quote is parsed as JavaScript inside
the KWin scripting engine rather than as a string comparison. Right
now `address` is sourced from KWin's own internalId (via the window
list plumbing in Windows.qml / Dock.qml / WindowSwitcherItem.qml), so
it's unlikely to contain a quote today — but nothing enforces that,
and the failure mode if it ever does is either a silently-broken
window action or unintended script execution in KWin's context.

### Confirmation

WinTone01:
Closes #270.

All nine window action methods interpolate `address` into a double-quoted JS string literal in the generated KWin script. A quote in the address ends the literal and the rest is parsed as JavaScript by KWin scripting engine. Addresses come from KWin own `internalId` today, so this is not reachable right now, but nothing enforces it.

Adds `escapeJsString()` and routes the address through it.

The methods that also interpolate numbers or booleans moved to `QString::arg` multi-argument overload. That is not cosmetic: with the chained form, `.arg(address).arg(x).arg(y)`, an address containing `%2` gets substituted by the next `.arg()` call. I checked it against the old form — `x%2y` with `setMaximize(false, false)` came out as `id==="xfalsey"`, a silently wrong comparison. The multi-arg overload substitutes in one pass, so the address is left alone.

Verified the normal case is byte-identical to the old output, so no behaviour change for real addresses.

### DIFF
--- a/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
+++ b/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
@@ -14,6 +14,22 @@
 
 namespace caelestia::services {
 
+namespace {
+
+// Window addresses are interpolated into double-quoted JS string literals in
+// the generated KWin scripts. They come from KWin's own internalId today, but
+// nothing enforces that, so escape them rather than trusting the source.
+QString escapeJsString(const QString& value) {
+    QString escaped = value;
+    escaped.replace('\\', QStringLiteral("\\\\"));
+    escaped.replace('"', QStringLiteral("\\\""));
+    escaped.replace('\n', QStringLiteral("\\n"));
+    escaped.replace('\r', QStringLiteral("\\r"));
+    return escaped;
+}
+
+} // namespace
+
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
LGTM!

PR #288, opened 2026-07-28
fix: don't crash when cliphist list crashes

### LINKED ISSUE #268: Crash risk: ClipboardManager::reload() can null-deref m_listProc if cliphist crashes
### Before you submit

### Diagnostics

Suggested fix: add the same `err == QProcess::FailedToStart` guard to
the errorOccurred handler in reload(), and/or null-check m_listProc
before calling deleteLater() in the finished handler.

### Issue details

In Services/clipboardmanager.cpp, reload() connects both
QProcess::finished and QProcess::errorOccurred on m_listProc, and
both handlers unconditionally run:

    m_listProc->deleteLater();
    m_listProc = nullptr;

When the cliphist child process crashes (as opposed to failing to
start), Qt emits errorOccurred(Crashed) followed by finished(...,
CrashExit) for the same QProcess instance. The errorOccurred handler
runs first, deletes and nulls m_listProc; then the finished handler
runs on the same object and calls m_listProc->deleteLater() again —
except m_listProc is now nullptr, causing a null pointer dereference.

This is inconsistent with clearHistory() in the same file, which
correctly guards this case:

    if (err == QProcess::FailedToStart && m_wipeProc) { ... }

FailedToStart is the only QProcess error where finished() is
guaranteed not to also fire, which is presumably why that guard
exists there. reload()'s handler for m_listProc is missing the
equivalent guard.

Repro conditions: anything that makes `cliphist list` crash rather
than exit normally (corrupted cliphist database, OOM-kill, session
teardown mid-request). Low frequency but a hard crash of the shell
process when it happens.

### Confirmation

WinTone01:
Closes #268.

`reload()` read `m_listProc` from both its `finished()` and `errorOccurred()` handlers and unconditionally did `deleteLater()` + `= nullptr`. A crashed cliphist emits `errorOccurred(Crashed)` and then `finished(CrashExit)` for the same process, so the second handler dereferences a null `m_listProc` and takes the shell down.

Reproduced it with a stub `cliphist` on PATH that raises SIGSEGV: before the change the harness exits 139, after it exits 0.

There is a second bug in the same place that the issue did not mention. The handlers read the member rather than the process they belong to, so if the `waitForFinished(200)` at the top of `reload()` times out, the old process `finished()` arrives *after* `m_listProc` has been reassigned and deletes the new process instead. Both handlers now capture the `QProcess*` and only clear the member when it still points at that instance.

`errorOccurred()` now only cleans up `FailedToStart`, the one error for which `finished()` is not also emitted — the same guard `clearHistory()` already uses. `finished()` also treats `CrashExit` as failure instead of trusting the exit code, which is meaningless on a crash.

### DIFF
--- a/shell/plugin/src/Caelestia/Services/clipboardmanager.cpp
+++ b/shell/plugin/src/Caelestia/Services/clipboardmanager.cpp
@@ -31,23 +31,38 @@ void ClipboardManager::reload() {
         m_listProc->waitForFinished(200);
     }
 
-    m_listProc = new QProcess(this);
-    m_listProc->setProgram("cliphist");
-    m_listProc->setArguments({"list"});
+    auto* proc = new QProcess(this);
+    m_listProc = proc;
+    proc->setProgram("cliphist");
+    proc->setArguments({"list"});
 
-    connect(m_listProc, &QProcess::finished, this, [this](int exitCode, QProcess::ExitStatus) {
-        if (exitCode != 0) {
-            qCWarning(lcClipboard) << "cliphist list failed with exit code" << exitCode;
-            m_items.clear();
-            emit itemsChanged();
-            m_listProc->deleteLater();
+    // Capture the process itself rather than reading m_listProc from the
+    // handlers: a crashed cliphist emits errorOccurred() *and* finished() for
+    // the same instance, and a superseded reload can deliver signals after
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
Great fixes, thank you!

PR #289, opened 2026-07-28
fix: pick up scheme.json when it's created after startup

### LINKED ISSUE #271: SchemeLoader never starts watching scheme.json if it doesn't exist at startup
### Before you submit

### Diagnostics

Suggested fix: also watch the parent directory
(QFileSystemWatcher::addPath on the containing folder) and, on
directoryChanged, check whether m_schemeStatePath now exists and add
it to the watcher if so.

### Issue details

SchemeLoader's constructor only calls m_watcher->addPath(m_schemeStatePath)
if QFile::exists(m_schemeStatePath) is already true at construction
time. There's no watch on the parent directory as a fallback, and the
fileChanged handler only re-arms watching for a path that's already
being watched — it can't pick up a file that doesn't exist yet.

Effect: on first run, before scheme.json has ever been written (fresh
install, first login, or a user who hasn't run `caelestia scheme set`
yet), SchemeLoader silently never starts watching it. The first
`caelestia scheme set` after that point won't trigger a live update —
the shell needs a full manual restart to notice.

### Confirmation

WinTone01:
Closes #271.

`SchemeLoader` only calls `addPath(m_schemeStatePath)` if the file already exists when the constructor runs, and the `fileChanged` handler can only re-arm a path that is already watched. On a fresh install, or a first login before `caelestia scheme set` has ever run, the file does not exist yet, so the watch is never established and the first scheme change needs a full shell restart to show up.

Now the state directory is watched as well, and `directoryChanged` starts watching the file once it appears.

Measured it with a small harness that constructs `SchemeLoader` against an empty state dir, then creates and updates scheme.json: before the change it emitted `currentSchemeChanged` zero times, after it caught all three (creation by the `caelestia scheme list` call, then both writes).

`watchSchemeState()` is idempotent, so the extra `directoryChanged` traffic from unrelated files in the directory costs a `QStringList::contains` and does not re-trigger a load.

### DIFF
--- a/shell/plugin/src/Caelestia/Services/schemeloader.cpp
+++ b/shell/plugin/src/Caelestia/Services/schemeloader.cpp
@@ -5,6 +5,7 @@
 
 #include <qdir.h>
 #include <qfile.h>
+#include <qfileinfo.h>
 #include <qfilesystemwatcher.h>
 #include <qjsondocument.h>
 #include <qjsonobject.h>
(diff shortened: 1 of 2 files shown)


0xSolanaceae (approved):
LGTM

PR #290, opened 2026-07-28
fix: build the keybind cheatsheet from live shortcuts, not a regex

### LINKED ISSUE #278: Keybind cheatsheet should read shortcut data from QML properties, not regex-scrape Shortcuts.qml
### Proposal

KeybindsModel::load() currently opens Shortcuts.qml as plain
text and regex-matches name:/key:/description: string literals per
line, using manual brace counting to detect block boundaries. Any
CustomShortcut whose name/key/description is built from a variable,
ternary, or template expression instead of a bare string literal
won't match the regex and silently won't show up in the cheatsheet —
there's no warning, it just goes missing.

Better: expose the shortcut list as an actual Q_INVOKABLE/property
read from the live QML object tree (or have CustomShortcut register
itself into a shared registry on creation), so the cheatsheet reflects
whatever's actually bound, however it's expressed, instead of only
what matches a specific string-literal pattern.

### Use case

Anyone who customizes their Shortcuts.qml beyond copy-paste
literal strings (e.g. binding a key from a config variable, which is
exactly the kind of thing a "heavily customizable dotfiles" project
encourages) gets a keybind that works perfectly but is invisible in
their own cheatsheet, with no error to tell them why.

### Additional context

_No response_

WinTone01:
Closes #278.

`KeybindsModel::load()` opened `Shortcuts.qml` as text and regex-matched `name:`/`key:`/`description:` string literals, with manual brace counting for block boundaries. Anything expressed as something other than a bare literal did not match and silently went missing.

This is not hypothetical. All nine krohnkite shortcuts are written as

```qml
key: Config.general.krohnkiteEnabled ? "Meta+Up" : ""
```

so the regex never matched them and they have never shown up in the cheatsheet, even while bound and working. The workspace shortcuts hit the same wall, which is why `load()` ended with a hardcoded `Super + 1..0` loop guessing at what the Instantiator produces.

I measured the current output before changing anything: 28 rows, 18 real plus the 10 guessed workspace rows, out of 62 `CustomShortcut` blocks. 29 of those declare a key at all, 19 as a literal and 10 as an expression.

`CustomShortcut` now registers itself with the model and re-registers when its name, key or description changes. The model holds entries in declaration order and coalesces rebuilds into one per event loop turn, so the burst of registrations at startup costs a single rebuild.

Three things improve beyond the missing rows:

- The cheatsheet follows config changes. Toggling `krohnkiteEnabled` adds or removes those nine rows live instead of needing a restart.
- It lists the shortcuts actually running, rather than whatever `Shortcuts.qml` happened to be found at the searched path — those could differ.
- The hardcoded workspace loop is gone. Those rows now come from the Instantiator real keys, so they stay correct if someone rebinds them.

The public surface is untouched — `keybinds`, `initialized`, `load()`, `query()` and `loaded()` behave as before, so `Keybinds.qml` needed no changes.

Verified by driving the real QML to C++ path with stand-in shortcuts covering literals, ternaries and template literals: ternary and template keys now appear with their evaluated values, `workspace10` correctly resolves to `Super + 0`, `"Meta+Space; Meta"` still collapses to `Super + Space`, and an empty key stays out of the list. Flipping the ternary condition took the list from 15 rows to 12 on the next turn. qmllint on `CustomShortcut.qml` is clean.

### DIFF
--- a/shell/components/misc/CustomShortcut.qml
+++ b/shell/components/misc/CustomShortcut.qml
@@ -15,6 +15,20 @@ Loader {
 
     active: true
 
+    // Report to the keybind cheatsheet directly. Doing it here means a
+    // shortcut whose name or key comes from a variable, ternary or template
+    // literal is listed exactly like one written as a plain string.
+    function publishKeybind(): void {
+        Caelestia.KeybindsModel.registerKeybind(root, root.name, root.key, root.description);
+    }
+
+    onNameChanged: publishKeybind()
+    onKeyChanged: publishKeybind()
+    onDescriptionChanged: publishKeybind()
+
+    Component.onCompleted: publishKeybind()
+    Component.onDestruction: Caelestia.KeybindsModel.unregisterKeybind(root)
+
     sourceComponent: Quickshell.env("HYPRLAND_INSTANCE_SIGNATURE") ? hyprShortcut : kdeShortcut
 
     Component {
--- a/shell/plugin/src/Caelestia/Services/keybindsmodel.cpp
+++ b/shell/plugin/src/Caelestia/Services/keybindsmodel.cpp
@@ -1,25 +1,38 @@
 // SPDX-License-Identifier: GPL-3.0-only
 #include "keybindsmodel.hpp"
 
-#include <QFile>
-#include <QRegularExpression>
-#include <QTextStream>
-#include <qdir.h>
-#include <qjsonarray.h>
-#include <qjsonobject.h>
 #include <qloggingcategory.h>
-#include <qprocess.h>
-#include <qsettings.h>
-#include <qstandardpaths.h>
 
 Q_LOGGING_CATEGORY(lcKeybinds, "caelestia.services.keybindsmodel", QtInfoMsg)
 
 namespace caelestia::services {
 
-KeybindsModel::KeybindsModel(QObject* parent)
-    : QObject(parent) {
-    load();
+namespace {
+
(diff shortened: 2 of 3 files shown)


ladybug-me:
Please checkout [ShortcutManager PR#266](https://github.com/ladybug-me/caelestia-dots-kde/pull/266), this has been already implemented there.
If I missed something, let me know.

0xSolanaceae (changes requested):
Adjust to fit new shortcut manager

WinTone01:
> Please checkout [ShortcutManager PR#266](https://github.com/ladybug-me/caelestia-dots-kde/pull/266), this has been already implemented there. If I missed something, let me know.

Oopsie :laughing:

WinTone01:
Superseded by #266 — closing.

I went back through the current dev. KeybindsModel has been rewritten as a QAbstractListModel over keybindsdefaults.hpp and KGlobalAccel, so the regex and brace counting this PR was working around are gone entirely. The specific breakage it fixed is covered: the krohnkite binds that the regex could never match are now named defaults gated on krohnkiteEnabled, and workspace1..10 are real entries instead of the hardcoded Super+1..0 loop the old model guessed with.

#266 also does the registration properly via GlobalShortcutDispatcher's shortcutRegistered/shortcutUnregistered, which is the same idea as this PR but carried further, with editable binds and persistent overrides on top.

Nothing here is worth rebasing onto that. You didn't miss anything — I looked for a gap and did not find one.

PR #291, opened 2026-07-28
fix(arpc): prefer the focused window for custom-app presence

### LINKED ISSUE #276: Custom-app presence should prefer the focused window over first list match
### Proposal

For Priority 2 only, check KWinActiveWindowBridge.activeWindow
against arpcTargetWindows first, falling back to the current full-list
scan only if nothing focused matches. Leave Priority 1 (Steam) scanning
the full list as-is, since showing "still playing" while alt-tabbed
away is probably intentional there.

### Use case

Right now a background app that happens to match your regex
can override whatever you're actually looking at, purely based on
window-list ordering rather than what's on screen.

### Additional context

_No response_

WinTone01:
Closes #276.

Priority 2 took the first window in `KWinActiveWindowBridge.windowList` matching `arpcTargetWindows`, so a background app could own your presence purely because of window-list ordering.

The focused window is now matched first. The seeding happens before the existing loop, and the loop already guards on `topTargetClass === ""`, so the full-list scan is untouched and still runs whenever nothing focused matches — behaviour only changes in the case the issue describes. Priority 1 keeps scanning the whole list, as suggested.

Two details beyond the literal ask:

- The match index comes from the focused window as well, so the custom label from `arpcTargetWindowLabels` belongs to the pattern that actually matched rather than to whichever one the list scan found.
- `onActiveWindowChanged` is now connected to `updatePresence()`. Only `onWindowListChanged` was wired up, so alt-tabbing between two matching apps would not have refreshed the presence and the fix would have been mostly invisible in practice.

Verified the selection with the patched control flow over five cases. With targets `["^code$", "^firefox$"]` and a window list ordered firefox-then-code:

| focused | result |
| --- | --- |
| `code` | `code` (previously `firefox`) |
| `firefox` | `firefox` |
| `dolphin` (no match) | `firefox`, via the fallback scan |
| nothing focused | `firefox`, via the fallback scan |
| no targets configured | no match |

That exercises a copy of the selection logic rather than the live component — `DiscordRPC.qml` pulls in DiscordIpc, GlobalConfig, SysInfo, Colours and CUtils, so standing it up headless was not worth the scaffolding. qmllint on the real file is clean.

### DIFF
--- a/shell/services/DiscordRPC.qml
+++ b/shell/services/DiscordRPC.qml
@@ -97,6 +97,7 @@ Item {
         enabled: root.active
         ignoreUnknownSignals: true
         function onWindowListChanged() { root.updatePresence(); }
+        function onActiveWindowChanged() { root.updatePresence(); }
     }
 
     Connections {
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
LGTM!

PR #292, opened 2026-07-28
feat(clipboard): tell the user when cliphist is missing

### LINKED ISSUE #269: Surface "cliphist not installed" state to the UI instead of only logging it
### Proposal

ClipboardManager currently has no way to tell QML that the cliphist
binary is missing or unusable — if QProcess::start() fails, the only
feedback is a qCWarning to the log. The launcher's clipboard view just
stays silently empty with no explanation to the user.

### Use case

Users who haven't installed cliphist (it's an external dependency,
not bundled) get a blank clipboard history with no indication of why,
which looks like a bug rather than a missing dependency. A simple
`Q_PROPERTY(bool available READ available NOTIFY availableChanged)`
set to false on FailedToStart, with a corresponding hint in the QML
view ("cliphist not found — install it to enable clipboard history"),
would make this self-explanatory instead of silently broken.

### Additional context

_No response_

WinTone01:
Closes #269.

**Stacked on #288** — the first commit here is that PR, since both touch `reload()`'s `errorOccurred` handler. Review the second commit, or merge #288 first and this rebases to a single commit.

Without cliphist installed the only feedback was a `qCWarning` in the log and a blank clipboard list, which reads as a bug rather than a missing optional dependency.

`ClipboardManager` now exposes `available`, driven by `QProcess::started` and `FailedToStart`. Those answer exactly the question the property asks — whether the binary could be launched — unlike the exit code, which conflates "not installed" with "cliphist ran and failed".

The launcher empty state gains a matching case, following the shape the wallpapers state already uses to tell an empty collection apart from an unmatched search:

> **cliphist not found**
> Install cliphist to enable clipboard history

The clipboard list is a mode of the app list rather than a `ContentList` state of its own, so the condition reads `root.currentList?.state === "clipboard"`.

Verified by running `reload()` under a PATH with no cliphist — `available` goes true to false with one `availableChanged` — and again with it present, where it stays true and emits nothing. Since `started()` drives the true case, installing cliphist while the shell is running clears the hint on the next reload rather than needing a restart. Build and qmllint are clean.

### DIFF
--- a/shell/modules/launcher/ContentList.qml
+++ b/shell/modules/launcher/ContentList.qml
@@ -346,6 +346,10 @@ Item {
     Row {
         id: empty
 
+        /// The clipboard list is a mode of the app list rather than a state of
+        /// its own, so ask the list itself.
+        readonly property bool cliphistMissing: root.currentList?.state === "clipboard" && !Clipboard.available
+
         opacity: root.currentList?.count === 0 ? 1 : 0
         scale: root.currentList?.count === 0 ? 1 : 0.5
 
@@ -357,6 +361,8 @@ Item {
 
         MaterialIcon {
             text: {
+                if (empty.cliphistMissing)
+                    return "content_paste_off";
                 if (root.state === "wallpapers")
                     return "wallpaper_slideshow";
                 if (root.state === "keybinds")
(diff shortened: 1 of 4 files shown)


0xSolanaceae (approved):
LGTM! Thank you again for your time helping contribute to the codebase!

PR #321, opened 2026-07-29
feat: automatic light/dark scheme switching on a schedule

### LINKED ISSUE #277: Time-based automatic scheme/variant switching (sunrise/sunset or fixed times)
### Proposal

An optional scheduler (fixed HH:MM times, or geo-based
sunrise/sunset like gammastep/redshift use) that calls the equivalent
of `caelestia scheme set` with the dark/light variant automatically.
SchemeLoader's file-watch plumbing already reacts to scheme.json
changes, so this is really just a small timer service that writes to
the same place, not a new reactive pathway.

### Use case

Every major theming tool (macOS, GNOME, Windows) ships
automatic dark-mode-at-night by now; right now it's a fully manual
switch here.

### Additional context

_No response_

WinTone01:
Closes #277.

Optional scheduler that flips the theme mode at fixed times or at local sunrise and sunset. Solar times reuse the coordinates already in `weatherLocation`, so there is no second location setting to keep in sync. The fixed times double as the fallback when solar times are unavailable — no location set, or polar day/night — so enabling this without a location still does something sensible.

Two decisions worth flagging:

- The check runs off the ticking clock, not a one-shot timer armed for the next boundary. A timer set eight hours ahead does not survive suspend/resume or a system clock change, which is what a laptop hits every night.
- It only acts on a boundary crossing. A manual mode switch stands until the next real sunrise or sunset instead of being reverted a minute later, and `setMode()` is skipped when the mode is already correct, so there is no redundant `caelestia scheme set` and no spurious notification.

The time arithmetic sits in `utils/scripts/solartime.js`, following the existing pattern for pure logic (see `Searcher.qml`), which keeps it free of config and service dependencies.

Checked the solar maths against known values, all within a minute:

| location / date | computed | published |
| --- | --- | --- |
| London 2026-06-21 | 03:44 / 20:22 UTC | 03:43 / 20:21 |
| London 2026-12-21 | 08:04 / 15:53 UTC | 08:04 / 15:53 |
| Sydney 2026-06-21 | 07:00 / 16:54 AEST | 07:00 / 16:54 |
| Tromso Jun and Dec | no sunrise/sunset | polar day / night |

Also covered `parseTime` edge cases (`24:00`, `12:60`, `0700` rejected), `parseCoords` range checks, and a full-day simulation confirming exactly two transitions in a normal window and correct behaviour in one wrapping past midnight. In that simulation a manual switch at 12:00 was left alone and produced no redundant call at the 19:00 boundary.

`shell.qml` needed `_autoSchemeInit`. Quickshell singletons are lazy, so without it the scheduler would only have run once someone opened the settings page.

Scope limit: the config takes `HH:MM` but the UI steps in whole hours. The stepper preserves minutes set by hand, so `07:30` in the config keeps its `:30` when the hour is changed. A minute-level picker would have meant changing the shared `CustomSpinBox`.

### DIFF
--- a/shell/modules/nexus/pages/ServicesPage.qml
+++ b/shell/modules/nexus/pages/ServicesPage.qml
@@ -1,3 +1,5 @@
+import "../../../utils/scripts/solartime.js" as Solar
+
 import QtQuick
 import QtQuick.Layouts
 import Quickshell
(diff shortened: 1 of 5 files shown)


0xSolanaceae (approved):
Great effort, thanks for the help!

PR #323, opened 2026-07-29
feat(arpc): clear the Discord presence while the user is idle

### LINKED ISSUE #274: Auto-clear or fade Discord presence after user idle timeout
### Proposal

Track idle time (KWin/Wayland idle protocol, or a simple
"no window focus change for N minutes" heuristic since
KWinActiveWindowBridge already tracks focus changes) and call
DiscordIpc.clearActivity() after a configurable idle threshold,
restoring presence on the next updatePresence() trigger once active
again.

### Use case

Rich presence is supposed to reflect what you're doing right
now, not what you were doing before you went to make coffee. Most
Discord RPC integrations people are used to (Spotify, native game
overlays) already do this.

### Additional context

_No response_

WinTone01:
Closes #274.

Rich presence is meant to say what you are doing now, not what you were doing before you walked away.

Uses Quickshell's `IdleMonitor`, i.e. the Wayland idle protocol, rather than the "no focus change for N minutes" heuristic the issue offers as a fallback. Focus changes miss the common case of sitting and reading, and the protocol is already available here.

`updatePresence()` returns early while idle. Without that, any of its other triggers — a window list change, a colour scheme change, a Steam fetch completing — would quietly put the presence back a moment after it was cleared.

Turning the feature off while idle resets the flag, so the presence is restored rather than stranded in the cleared state.

The timeout defaults to 0, which disables it. Anyone already using ARPC keeps exactly the behaviour they have now until they opt in from the settings page. The issue notes that most integrations do this by default, so an on-by-default is arguable — I went the other way so nobody's presence quietly disappears after an update. Easy to flip if you disagree.

Walked the state machine through seven steps: presence sent while active, cleared on going idle, skipped for window-list and colour changes while away, sent again on return, and — the one that would otherwise have stranded it — cleared while idle then correctly restored when the timeout is set to 0 from the settings page.

### DIFF
--- a/shell/modules/nexus/pages/services/ArpcPage.qml
+++ b/shell/modules/nexus/pages/services/ArpcPage.qml
@@ -73,11 +73,23 @@ PageBase {
         ToggleRow {
             Layout.topMargin: Tokens.spacing.extraSmall / 2 - parent.spacing
             Layout.fillWidth: true
-            last: true
             text: qsTr("Broadcast Caelestia info")
             subtext: qsTr("Broadcast shell uptime and system info")
             checked: GlobalConfig.services.arpcCaelestiaInfo
             onToggled: GlobalConfig.services.arpcCaelestiaInfo = checked
+        }
+
+        StepperRow {
+            Layout.topMargin: Tokens.spacing.extraSmall / 2 - parent.spacing
+            Layout.fillWidth: true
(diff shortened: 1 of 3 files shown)


0xSolanaceae (approved):
LGTM!

PR #328, opened 2026-07-29
refactor: improve clipboard cache path resolution and enforce strict …

### LINKED ISSUE #312: `ClipboardManager` writes decoded clipboard images to `/tmp/caelestia-clipboard/` with default permissions when `XDG_RUNTIME_DIR` is unset
### What happened?

`shell/plugin/src/Caelestia/Services/clipboardmanager.cpp`, constructor:
```cpp
const auto runtimeDir = qEnvironmentVariable("XDG_RUNTIME_DIR", "/tmp");
m_imageCacheDir = runtimeDir + "/caelestia-clipboard";
```
When `XDG_RUNTIME_DIR` is set (the normal case under a systemd-managed graphical
session), this resolves to `/run/user/<uid>/caelestia-clipboard`, and it's protected
by inheritance since `/run/user/<uid>` itself is created mode `0700` by
`systemd-logind` — fine, regardless of what mode the subdirectory itself ends up
with, because nobody else can even traverse into `/run/user/<uid>` to reach it.

The fallback is where this breaks: if `XDG_RUNTIME_DIR` is unset — which the code
explicitly anticipates by providing `/tmp` as a fallback default, implying it's a
real possibility the author is guarding for (non-systemd sessions, some display
managers, or the shell launched manually outside a full login) — `m_imageCacheDir`
becomes `/tmp/caelestia-clipboard`. `/tmp` is shared and world-traversable. The
directory is created via `QDir().mkpath(m_imageCacheDir)` (line 92, and again in
`decodeImage()` and `clearHistory()`) with no explicit mode set, so it's created
using the process's default umask — typically `0755` — and each decoded PNG is
written via a plain `QFile::open(QIODevice::WriteOnly)` (in `decodeImage()`), again
with default permissions (typically `0644`). Both directory and files end up
world-readable.

Net effect on that fallback path: any other local user on the machine can `ls
/tmp/caelestia-clipboard/` and read every image that's passed through the
clipboard, for as long as it stays in `cliphist`'s history.

Expected: clipboard image cache should never land somewhere other users can read
it, regardless of which fallback path gets taken.

### Steps to reproduce

1. Launch a session (or just the shell manually) with `XDG_RUNTIME_DIR` unset: `env
   -u XDG_RUNTIME_DIR quickshell -c caelestia/shell.qml` (or however the shell is
   normally invoked, with that variable stripped from the environment first).
2. Copy an image to the clipboard so `cliphist` records it, then open the clipboard
   history in the launcher so `ClipboardManager::decodeImage()` runs.
3. As a second local user on the same machine, `ls -la /tmp/caelestia-clipboard/` —
   the directory and its `.png` files are both readable (`ls -ld` shows
   `drwxr-xr-x` on the directory, `-rw-r--r--` on the files).

### Root cause
`shell/plugin/src/Caelestia/Services/clipboardmanager.cpp`: the `/tmp` fallback in
the constructor (line 19), combined with `QDir().mkpath()` (lines 92, 120) and
`QFile::open(WriteOnly)` (line 140) never setting an explicit restrictive mode
either for the directory or the files written into it.

### Suggested fix direction
After `mkpath()`, explicitly tighten permissions with
`QFile::setPermissions(m_imageCacheDir, QFile::ReadOwner | QFile::WriteOwner |
QFile::ExeOwner)` (i.e. `0700`) whenever the directory is created, and set `0600` on
each decoded PNG right after writing it — cheap, and makes the cache safe regardless
of which parent directory it ends up under. Even better: skip the `/tmp` fallback
altogether and fail loudly (or fall back to
`QStandardPaths::writableLocation(QStandardPaths::TempLocation)` combined with a
per-user, explicitly-`0700` subdirectory) rather than silently degrading into a
shared location.

### Environment

```markdown

```

### Logs or screenshots

_No response_

ladybug-me:
…directory permissions

## Type

- [ ] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

### DIFF
--- a/shell/plugin/src/Caelestia/Services/clipboardmanager.cpp
+++ b/shell/plugin/src/Caelestia/Services/clipboardmanager.cpp
@@ -9,6 +9,7 @@
 #include <qfileinfo.h>
 #include <qloggingcategory.h>
 #include <qregularexpression.h>
+#include <QStandardPaths>
 
 Q_LOGGING_CATEGORY(lcClipboard, "caelestia.services.clipboard", QtInfoMsg)
 
@@ -16,8 +17,11 @@ namespace caelestia::services {
 
 ClipboardManager::ClipboardManager(QObject* parent)
     : QObject(parent) {
-    const auto runtimeDir = qEnvironmentVariable("XDG_RUNTIME_DIR", "/tmp");
-    m_imageCacheDir = runtimeDir + "/caelestia-clipboard";
+    QString runtimeDir = QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation);
+    if (runtimeDir.isEmpty()) {
+        runtimeDir = QStandardPaths::writableLocation(QStandardPaths::TempLocation) + "/caelestia-" + qEnvironmentVariable("USER");
+    }
+    m_imageCacheDir = runtimeDir + "/clipboard";
 }
 
 QVariantList ClipboardManager::items() const { return m_items; }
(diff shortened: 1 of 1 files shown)


0xSolanaceae (approved):
LGTM!

PR #385, opened 2026-07-31
fix: guard Game Mode's KWin save step so a reload can't corrupt it

### LINKED ISSUE #370: Game Mode's saved "previous KWin settings" file is not idempotent, so a shell reload while Game Mode is active permanently loses the user's real blur/animation settings
### Issue type

### What happened?

On KWin, `GameMode.qml`'s `onEnabledChanged` handler calls `applyKwin(true)` when Game Mode turns on. That shell-out reads the user's *current* `blurEnabled`/`AnimationDurationFactor` via `kreadconfig6` and writes them to `$HOME/.cache/caelestia/gamemode-prev`, then disables blur/animations. `applyKwin(false)` (on Game Mode turning off) reads that same file back and restores those values. This round-trip is only correct if `applyKwin(true)` runs exactly once per Game Mode session — but `props.enabled` is a `PersistentProperties` value (`reloadableId: "gameMode"`), the same mechanism this codebase uses everywhere to survive a Quickshell hot-reload (`qs -c caelestia reload`) or a crash/restart. If the shell reloads while Game Mode is already on, the persisted `enabled: true` gets restored into the freshly-constructed singleton, which is a genuine property write and fires `onEnabledChanged` again with `enabled === true`. `applyKwin(true)` runs a second time — except now `kreadconfig6` reads back `blurEnabled=false` / `AnimationDurationFactor=0`, i.e. Game Mode's *own* already-applied values, and overwrites `gamemode-prev` with those. The original pre-Game-Mode settings are gone. When the user later actually turns Game Mode off, `applyKwin(false)` restores the corrupted file's contents — `blurEnabled=false`, `AnimationDurationFactor=0` — leaving blur and animations permanently disabled system-wide until the user manually re-enables them in System Settings, with no indication anything went wrong.

### Steps to reproduce

1. Confirm blur and animations are enabled in System Settings (KWin).
2. Enable Game Mode (via IPC `caelestia gameMode enable` or the UI) — confirm `~/.cache/caelestia/gamemode-prev` now contains `true` / `1` (or whatever the real prior values were).
3. While Game Mode is still on, trigger a Quickshell reload (`qs -c caelestia reload`, or crash/restart the shell process).
4. Check `~/.cache/caelestia/gamemode-prev` again — it now contains `false` / `0`, Game Mode's own values, not the originals.
5. Turn Game Mode off. Blur and animations stay disabled; System Settings shows them off even though the user never turned them off themselves.

### Caelestia version

kwin_port branch, `shell/services/GameMode.qml`

### Distro

Any

### Logs or screenshots

N/A — root cause is `applyKwin(true)`'s save step reading live `kwinrc`/`kdeglobals` state unconditionally on every invocation instead of only the first time a Game Mode session begins (e.g. guard on whether `gamemode-prev` already exists for this session, or track "already saved" in the same `PersistentProperties` block).

### LINKED ISSUE #371: Guard Game Mode's "save previous settings" step so it only fires once per session, and expose it as a general pattern for other stateful toggles
### What's the idea?

Add a `PersistentProperties`-backed `_prevSaved: bool` flag to `GameMode.qml`, set once inside `applyKwin(true)` and checked before the `kreadconfig6` save step runs again, so a reload/restart while Game Mode is active reuses the already-saved previous values instead of re-capturing (and corrupting) them from the current, already-modified state. The same "toggle-something, remember what it was before, put it back later" pattern shows up anywhere a feature temporarily overrides a KDE setting — worth documenting as the standard approach so future toggle-based features don't hit the same reload-reentrancy trap.

### Why would it be useful?

This directly fixes Bug Game Mode's saved "previous KWin settings" file is not idempotent, so a shell reload while Game Mode is active permanently loses the user's real blur/animation settings
above, and more generally protects any current or future stateful KWin/KDE override feature from losing user settings across the exact kind of hot-reload this codebase relies on `PersistentProperties`/`reloadableId` for everywhere else. Right now the fix has to be rediscovered per-feature; a shared save-once helper (or at least a documented convention) prevents the same class of bug from recurring as more of these toggles get added.

WinTone01:
Closes #370, closes #371.

`applyKwin(true)` reads the user's current `blurEnabled`/`AnimationDurationFactor` and saves them to `gamemode-prev` before switching them off, so `applyKwin(false)` can put them back. That round-trip only holds if `applyKwin(true)` runs exactly once per game mode session - but `props.enabled` is `PersistentProperties`-backed (`reloadableId: "gameMode"`), the mechanism this codebase uses everywhere to survive a Quickshell reload. If the shell reloads while game mode is already on, the persisted `enabled: true` gets restored into the freshly-constructed singleton, firing `onEnabledChanged` again with `enabled === true`. The save step would run a second time - except by then `kreadconfig6` reads back game mode's own already-applied values (blur off, animations off) and overwrites `gamemode-prev` with those, permanently losing the user's real settings.

`props._prevSaved`, on the same `PersistentProperties` object as `enabled`, records that the save already happened for this session. It is restored alongside `enabled` by the same mechanism, so it survives the exact reload that would otherwise retrigger the save, and is reset when game mode turns off so the next session saves correctly. `applyKwin()` only builds the save half of the script when the flag isn't already set.

Per #371, also generalises the shape as a comment on `applyKwin()` - any feature that temporarily overrides a KDE/Hyprland setting and remembers the prior value to restore later needs the same guard, for the same reason. No shared helper was added since Game Mode is currently the only feature doing this; the comment documents the pattern for whoever writes the next one.

## Testing

Verified the normal single-session cycle end to end on KDE Plasma 6.7 / kwin_wayland: enabling saves the live values and applies game mode's, disabling restores the saved values - no regression there.

I could not reproduce the reload trigger itself in this environment: `caelestia shell -k` + relaunch does not preserve `PersistentProperties` here (confirmed with `idleInhibitor` too, as a control - it also resets on a plain process restart), and this quickshell build (0.3.0-git) exposes no `reload` subcommand and did not visibly hot-reload on a watched file being touched or rewritten. The fix is verified by code review instead: `_prevSaved` lives in the same `PersistentProperties` block as `enabled`, so whatever restores `enabled: true` on a singleton reconstruction restores `_prevSaved` right alongside it, atomically - which is the guarantee the guard depends on. If someone can trigger a real hot reload on their setup, re-running the repro steps from #370 with this branch would be good extra confirmation.

### DIFF
--- a/shell/services/GameMode.qml
+++ b/shell/services/GameMode.qml
@@ -91,21 +91,47 @@ Singleton {
     // KWin's equivalents of the Hyprland options above: window animations and the
     // blur effect. Both are read back before being changed so a user who already
     // had them off does not get them switched on when game mode ends.
+    //
+    // The read-back has to happen at most once per game mode session, not once
+    // per call. props.enabled is PersistentProperties-backed (reloadableId
+    // "gameMode"), the same mechanism this codebase uses everywhere to survive a
+    // Quickshell hot-reload — so a reload while game mode is already on restores
+    // enabled: true into the freshly-constructed singleton, which fires
+    // onEnabledChanged again with enabled === true. Without a guard, applyKwin(true)
+    // would run its save step a second time, except by then kreadconfig6 reads back
+    // game mode's *own* already-applied values (blur off, animations off) and
+    // overwrites gamemode-prev with those — the user's real settings are gone, and
+    // applyKwin(false) later "restores" them to off. props._prevSaved records that
+    // the save already happened for this session; being on the same
+    // PersistentProperties object, it survives the reload that would otherwise
+    // retrigger the save.
+    //
+    // The general shape — toggle something, remember what it was before, put it
+    // back later — recurs for any feature that temporarily overrides a KDE/Hyprland
+    // setting. Guard the save step the same way: a PersistentProperties flag, set
+    // once when the override begins and cleared only when it ends, rather than
+    // assuming the "on" handler only ever runs once per session.
     function applyKwin(enable: bool): void {
-        const script = enable
-            ? 'prevBlur="$(kreadconfig6 --file kwinrc --group Plugins --key blurEnabled --default true)"; ' +
-              'prevAnim="$(kreadconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor --default 1)"; ' +
-              'mkdir -p "$HOME/.cache/caelestia"; printf "%s\\n%s\\n" "$prevBlur" "$prevAnim" > "$HOME/.cache/caelestia/gamemode-prev"; ' +
-              'kwriteconfig6 --file kwinrc --group Plugins --key blurEnabled false; ' +
-              'kwriteconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor 0; ' +
-              'qdbus6 org.kde.KWin /KWin reconfigure >/dev/null 2>&1'
-            : 'p="$HOME/.cache/caelestia/gamemode-prev"; ' +
-              'blur="$(sed -n 1p "$p" 2>/dev/null)"; anim="$(sed -n 2p "$p" 2>/dev/null)"; ' +
-              '[ -n "$blur" ] || blur=true; [ -n "$anim" ] || anim=1; ' +
-              'kwriteconfig6 --file kwinrc --group Plugins --key blurEnabled "$blur"; ' +
-              'kwriteconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor "$anim"; ' +
-              'qdbus6 org.kde.KWin /KWin reconfigure >/dev/null 2>&1';
-        Quickshell.execDetached(["sh", "-c", script]);
+        if (enable) {
+            const saveStep = props._prevSaved ? "" :
+                'prevBlur="$(kreadconfig6 --file kwinrc --group Plugins --key blurEnabled --default true)"; ' +
+                'prevAnim="$(kreadconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor --default 1)"; ' +
+                'mkdir -p "$HOME/.cache/caelestia"; printf "%s\\n%s\\n" "$prevBlur" "$prevAnim" > "$HOME/.cache/caelestia/gamemode-prev"; ';
+            props._prevSaved = true;
+            Quickshell.execDetached(["sh", "-c", saveStep +
+                'kwriteconfig6 --file kwinrc --group Plugins --key blurEnabled false; ' +
+                'kwriteconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor 0; ' +
+                'qdbus6 org.kde.KWin /KWin reconfigure >/dev/null 2>&1']);
+        } else {
+            props._prevSaved = false;
+            Quickshell.execDetached(["sh", "-c",
+                'p="$HOME/.cache/caelestia/gamemode-prev"; ' +
+                'blur="$(sed -n 1p "$p" 2>/dev/null)"; anim="$(sed -n 2p "$p" 2>/dev/null)"; ' +
+                '[ -n "$blur" ] || blur=true; [ -n "$anim" ] || anim=1; ' +
+                'kwriteconfig6 --file kwinrc --group Plugins --key blurEnabled "$blur"; ' +
+                'kwriteconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor "$anim"; ' +
+                'qdbus6 org.kde.KWin /KWin reconfigure >/dev/null 2>&1']);
+        }
     }
 
     onEnabledChanged: {
@@ -154,6 +180,12 @@ Singleton {
         // there. onConfigReloaded below re-applies the options on Hyprland, so
         // nothing needed the binding anyway.
         property bool enabled: false
+
+        // See applyKwin(): true once the previous KWin blur/animation state has
+        // been captured for the current game mode session, so a reload while
+        // still enabled does not re-capture (and corrupt) it from game mode's own
+        // already-applied values. Reset when game mode turns off.
+        property bool _prevSaved: false
 
         reloadableId: "gameMode"
     }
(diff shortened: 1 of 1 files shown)


ladybug-me:
Tested it
**The suggested issue still exists**
If shell crashes or user **restarts** the shell, the kde settings are still gone.

**Fix**: Use `~/.cache/caelestia/gamemode-prev` as a state file too.
Delete this file on game mode exit by `applykwin(false)`.

Now on shell startup, if it exists, Game mode was active and the old settings are there.
If not, game mode exited successfully.

ladybug-me:
Also, a little out of scope for this pr but the current qdbus6 commands are not able to reflect the changes in the live shell (although settings do change in kde settings)

If you can make it work, let me know

0xSolanaceae (changes requested):
See additional comments from @ladybug-me

WinTone01:
Pushed a fix using exactly the approach you suggested — thanks for testing and catching this, my original guard didn't cover the actual reported case.

`_prevSaved` is gone. `gamemode-prev`'s own existence on disk is now the guard: `applyKwin(true)` only captures and writes it if it doesn't already exist, `applyKwin(false)` deletes it after restoring. Neither depends on `PersistentProperties`, which I confirmed (with `idleInhibitor` as a control) does not reliably survive a real restart in this environment.

That still left the restart itself unhandled on its own — `enabled` resets to `false` regardless, so nothing would ever call `applyKwin(false)` to put the real settings back, exactly matching what you saw. Added a `FileView` that checks for `gamemode-prev` on startup and sets `enabled` back to `true` if it's there, since its existence alone means a previous session never got to turn itself off.

Ran your exact repro end to end (reading `kwinrc`/`kdeglobals` via `kreadconfig6` at each step):
1. Known-on baseline, enable game mode → prev file saves it, KWin goes off.
2. Kill + restart the shell process while still "on".
3. `isEnabled()` correctly comes back `true` post-restart, prev file unchanged.
4. Disabling afterward restores the original baseline and deletes the file.

Normal enable/disable with no restart in between still works as before.

On the qdbus6-not-reflecting-live-shell point — I don't have enough detail yet to know if that means the KWin compositor's actual live blur rendering, or Caelestia's own settings UI not reflecting an externally-changed value. Left it out of this PR as you said; happy to take a look if you can point me at which one.

ladybug-me:
Great work.

> On the qdbus6-not-reflecting-live-shell point — I don't have enough detail yet to know if that means the KWin compositor's actual live blur rendering, or Caelestia's own settings UI not reflecting an externally-changed value. Left it out of this PR as you said; happy to take a look if you can point me at which one.


It perfectly reflects changes in caelestia's own ui but doesn't reflect in KWin compositor's live rendering.

(Verified by animation slider in kde settings changing to 100% instantaneous but the effect is not observed in the live environment. Works after you change it manually in kde settings.)

WinTone01:
Found it — captured the real traffic with dbus-monitor while changing the animation speed slider through System Settings myself.

It broadcasts `org.kde.kconfig.notify`'s `ConfigChanged` signal on `/kdeglobals` with payload `{"KDE": ["AnimationDurationFactor"]}`. `kwriteconfig6` has a `--notify` flag that emits that exact same signal - confirmed byte-for-byte identical against the captured System Settings traffic. Pushed a fix using it for both the disable and restore writes.

`blurEnabled` is a KWin effect setting in `kwinrc`, not a generic Qt/Plasma one, so `reconfigure()` already covers it live - didn't touch that part.

Verified GameMode's own `applyKwin(true)` call now emits the identical `ConfigChanged` signal, and re-ran the full enable/disable cycle afterward with no regression.

0xSolanaceae (changes requested):
One more thing, can you make the script's cache path XDG-aware to match Paths.cache?

ladybug-me:
**Please change the file name from gamemode-prev to gamemode-state**

After testing, there was a bug where game mode will stay always active due to the presence of old file. (File from before implementing the `use it as a state file too`).
A simple fix is to change the file name.

shell/services/GameMode.qml
```diff
@@ -91,21 +92,59 @@ Singleton {
...
-            : 'p="$HOME/.cache/caelestia/gamemode-prev"; ' +
-              'blur="$(sed -n 1p "$p" 2>/dev/null)"; anim="$(sed -n 2p "$p" 2>/dev/null)"; ' +
-              '[ -n "$blur" ] || blur=true; [ -n "$anim" ] || anim=1; ' +
-              'kwriteconfig6 --file kwinrc --group Plugins --key blurEnabled "$blur"; ' +
-              'kwriteconfig6 --file kdeglobals --group KDE --key AnimationDurationFactor "$anim"; ' +
-              'qdbus6 org.kde.KWin /KWin reconfigure >/dev/null 2>&1';
-        Quickshell.execDetached(["sh", "-c", script]);
+        if (enable) {
+            Quickshell.execDetached(["sh", "-c",
+                'p="$HOME/.cache/caelestia/gamemode-prev"; ' +
```
ladybug-me: change this and the others to `gamemode-state`

ladybug-me (changes requested):

WinTone01:
Done — renamed to `gamemode-state` everywhere (save/restore paths and the startup FileView).

Reproduced the exact scenario before pushing: created a stale `gamemode-prev` file with no matching `gamemode-state`, restarted the shell, confirmed `isEnabled()` correctly comes back `false` instead of getting stuck on. Re-ran the full enable/disable cycle with the new filename afterward, no regressions.

WinTone01:
Missed this one earlier, thanks for the catch — pushed a fix.

`applyKwin()`'s shell script was hardcoding `$HOME/.cache/caelestia`. `Paths.cache` already resolves `XDG_CACHE_HOME` correctly with the same fallback, and the `FileView` added earlier in this PR already used it — the shell script was the one place still duplicating the fallback logic instead of sharing it, so it could have quietly diverged on a system with `XDG_CACHE_HOME` set. Both now build the state file path from the same `Paths.cache` value.

Verified by restarting the shell with `XDG_CACHE_HOME` pointed at a scratch directory (confirmed via the process's own environment that it actually took effect), then running the full enable/disable cycle - the state file was created and cleaned up there, and never touched the default `~/.cache/caelestia`. Restarted normally afterward to confirm the unset case still falls back correctly.

0xSolanaceae (approved):
thanks!

ladybug-me (approved):

PR #386, opened 2026-07-31
fix: sync the extracted first-frame thumbnail to Plasma for video wallpapers

### LINKED ISSUE #379: Video wallpapers sync to KDE's native desktop layer as a broken static image instead of the already-generated first-frame thumbnail
### Issue type

### What happened?

`Wallpapers.qml`'s `onActualCurrentChanged` handler unconditionally pushes whatever `actualCurrent` currently is straight into Plasma's real desktop wallpaper config over D-Bus, hardcoding the target plugin to the static-image renderer:

```js
d.wallpaperPlugin = "org.kde.image";
...
d.writeConfig("Image", "file://$1")  // $1 = actualCurrent
```

`actualCurrent` gets set from `setWallpaper(path)` (called for any wallpaper, video or image alike) with no branch on file type. When the current wallpaper is a video — this codebase clearly supports video wallpapers, with a dedicated `VideoWallpaperPlayer.qml` renderer and a `getThumbnailPath()`/`videoThumbs` first-frame-extraction cache specifically built because "video wallpapers have no still to show" (per this same file's own comment) — the raw video file path gets written into `org.kde.image`'s `Image` config key, which can only render static images. `org.kde.image` can't decode a video file, so Plasma's actual desktop wallpaper layer ends up blank/broken underneath whatever the current video wallpaper is. This doesn't show under normal operation because Caelestia's own background layer draws over the real Plasma desktop — but it's exactly what's exposed the moment that's not true: a shell crash or manual `qs -c caelestia` exit, briefly during Quickshell's own startup before its background window is up, or if someone temporarily switches to stock Plasma to troubleshoot. The fix is sitting right there in the same file: `getThumbnailPath()` already produces a cached still (`${Paths.cache}/wallpapers/${sha256(path)}/first_frame.png`) for exactly this "video has no still" problem in the picker UI — the Plasma sync in `onActualCurrentChanged` just never branches on `Images.isVideo(actualCurrent)` to use it.

### Steps to reproduce

1. Set a video file as the current wallpaper.
2. Confirm `${Paths.cache}/wallpapers/<hash>/first_frame.png` exists (the picker's thumbnail cache already generated it).
3. Query the live Plasma config directly: `qdbus6 org.kde.plasmashell /PlasmaShell org.kde.PlasmaShell.evaluateScript 'print(desktops()[0].readConfig("Image"))'` (adjusted per the wallpaper's actual config group) — it reports the raw video file path, not the still frame.
4. Exit Quickshell (`pkill quickshell` or the equivalent) so the real Plasma desktop is exposed — it shows a blank/broken wallpaper instead of at least the extracted still frame.

### Caelestia version

kwin_port branch, `shell/services/Wallpapers.qml`

### Distro

Any

### Logs or screenshots

N/A — confirmed by tracing `actualCurrent`'s only two write sites (`setWallpaper()` and the startup-state read) against `onActualCurrentChanged`'s D-Bus call, neither of which checks `Images.isVideo()` before handing the path to `org.kde.image`.

### LINKED ISSUE #380: Sync the extracted first-frame thumbnail to Plasma's native wallpaper config for video wallpapers instead of the raw video path
### What's the idea?

In `Wallpapers.qml`'s `onActualCurrentChanged`, branch on `Images.isVideo(actualCurrent)`: if true, wait for (or trigger) `getThumbnailPath(actualCurrent)`'s cached `first_frame.png` and pass that to the `d.writeConfig("Image", ...)` call instead of the raw video path; otherwise keep the current behavior unchanged.

### Why would it be useful?

This directly fixes Bug Video wallpapers sync to KDE's native desktop layer as a broken static image instead of the already-generated first-frame thumbnail with essentially no new code — the thumbnail extraction pipeline already exists and is already keyed exactly the way this call site would need. It means the real KDE desktop shows a sensible matching still frame instead of a blank/broken wallpaper any time Caelestia's own background layer isn't covering it (crash, manual exit, troubleshooting fallback to stock Plasma), which is a much better first impression/safety net than what's there now.

WinTone01:
Closes #379, closes #380.

`onActualCurrentChanged` pushed whatever the current wallpaper's raw path was straight into Plasma's real desktop config over D-Bus, hardcoded to `org.kde.image` - the static-image renderer. For a video wallpaper that path can't be decoded at all, so the actual Plasma desktop layer underneath ends up blank/broken. This is invisible under normal operation since Caelestia's own background layer draws over the real desktop, but it's exactly what's exposed the moment that's not true: a crash, a manual `qs -c caelestia` exit, or switching back to stock Plasma to troubleshoot.

`thumbFor()` already does the "video path in, still frame out" resolution the wallpaper pickers (`WallpaperItem.qml`, `WallItem.qml`) use - it returns the cached first-frame extraction if one exists, or `""` while kicking off the ffmpeg extraction if it doesn't. The Plasma sync now goes through that same function instead of writing the raw path unconditionally, and a new `onVideoThumbsChanged` handler re-syncs once a pending extraction lands. While nothing is ready yet, Plasma's previous (still valid) wallpaper is left alone rather than actively pointed at something it can't render - strictly better than the alternative of writing a path that's guaranteed broken.

## Testing

Live end to end on KDE Plasma 6.7 / kwin_wayland, reading Plasma's `Image` config directly via `qdbus6` after each step:

- **Image wallpaper**: unaffected, syncs immediately as before.
- **Video wallpaper, cold cache** (fresh shell process, no prior extraction): Plasma keeps showing the previous wallpaper for the ~1s the ffmpeg extraction takes, then picks up the finished still frame once it lands - the raw `.mp4` path never reaches Plasma's config at any point in between.
- **Video wallpaper, warm cache** (frame already extracted from an earlier session): syncs to the cached still immediately.

### DIFF
--- a/shell/services/Wallpapers.qml
+++ b/shell/services/Wallpapers.qml
@@ -123,7 +123,17 @@ Searcher {
 
     function setWallpaper(path: string): void {
         actualCurrent = path;
-        Quickshell.execDetached(["caelestia", "wallpaper", "-f", path, ...smartArg]);
+        if (Images.isVideo(path)) {
+            const thumb = thumbFor(path);
+            if (thumb !== "") {
+                const script = 'caelestia wallpaper -f "$1" ' + root.smartArg.join(" ") + '; printf "%s" "$2" > "$3"';
+                Quickshell.execDetached(["sh", "-c", script, "--", thumb, path, root.currentNamePath]);
+            } else {
+                Quickshell.execDetached(["sh", "-c", 'printf "%s" "$1" > "$2"', "--", path, root.currentNamePath]);
+            }
+        } else {
+            Quickshell.execDetached(["caelestia", "wallpaper", "-f", path, ...smartArg]);
+        }
     }
 
     function preview(path: string): void {
@@ -200,6 +210,10 @@ Searcher {
             const m = root.videoThumbs;
             m[path] = out;
             root.videoThumbs = Object.assign({}, m);   // a copy, so bindings re-run
+            if (path === root.actualCurrent) {
+                const script = 'caelestia wallpaper -f "$1" ' + root.smartArg.join(" ") + '; printf "%s" "$2" > "$3"';
+                Quickshell.execDetached(["sh", "-c", script, "--", out, path, root.currentNamePath]);
+            }
         }
         const pending = root.videoThumbsPending;
         delete pending[path];
@@ -246,6 +260,9 @@ Searcher {
             if (!wall) {
                 wall = root.fallback;
                 Quickshell.execDetached(["caelestia", "wallpaper", "-f", root.fallback, ...root.smartArg]);
+            }
+            if (Images.isVideo(root.actualCurrent) && wall === root.getThumbnailPath(root.actualCurrent)) {
+                return;
             }
             root.actualCurrent = wall;
             root.previewColourLock = false;
--- a/shell/services/Wallpapers.qml
+++ b/shell/services/Wallpapers.qml
@@ -23,9 +23,36 @@ Searcher {
     property bool previewColourLock
     property bool pendingPreviewClear
 
-    onActualCurrentChanged: {
-        // Sync KDE Plasma wallpaper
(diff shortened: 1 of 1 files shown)


ladybug-me:
Tested
Extracted frame quality is reduced but that is expected 
Added a workaround for saving video types in path.txt

PR #425, opened 2026-08-06
fix: stop refreshWindows() dropping the output field

WinTone01:
## What does this change?

`KWinActiveWindowBridge::refreshWindows()` rebuilds the window list from scratch and replaces `m_windowList` wholesale, but its payload was missing the `output` field that the event-driven `notifyWindowList()` sends.

It runs on every workspace change. So after the first switch, every window looked like it belonged to no screen, and the per-monitor `Hypr.hasFullscreenOn()` check could never match — the shell stopped getting out of the way of fullscreen windows until some unrelated event pushed a complete list again.

Its `maximized` field also read `w.maximized`, which is not a property of KWin's `Window`. The real state is `maximizeMode === 3`, as `notifyWindowList()` already had it, so every window was being reported as not maximized.

## How did you test it?

- Compared the two payload builders field by field; `output` was absent from one and `maximized` read a property that does not exist on `Window` (confirmed against the KWin scripting API).
- Dumped the live payload from `$XDG_RUNTIME_DIR/qs_kwin_windows.json` after workspace switches on a running shell: every entry now carries its `output` (`DP-1` here), where previously the field vanished after the first switch.
- Running in a live shell since, on a multi-monitor setup.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

Both payloads feed the same `m_windowList`, so they have to stay identical field for field — a field present in only one silently disappears from the shell's view whenever that path happens to run last. I added a comment saying so, but if you would rather this be structural than a comment, the two builders could share a single generator function instead. Happy to do that here if you prefer.

### DIFF
--- a/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
+++ b/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
@@ -53,6 +53,65 @@ void KWinActiveWindowBridgeAdaptor::notifyCurrentDesktop(int desktop) {
     }
 }
 
+// The single description of a window-list entry, shared by every path that
+// produces one.
+//
+// Both the persistent bridge script and the one-shot refresh replace the same
+// m_windowList wholesale, so a field present in only one of them silently
+// disappears from the shell's view whenever that path happens to run last —
+// which is exactly how the `output` field went missing and blinded the
+// per-monitor fullscreen check after a workspace switch. Interpolated into both
+// rather than restated, so they cannot drift again.
+static const QString kWindowListJs = R"js(
+function caelestiaDesktopIndex(w) {
+    if (!w.desktops || w.desktops.length === 0) return -1;
+    let d = w.desktops[0];
+    if (!d) return -1;
+    let allD = workspace.desktops;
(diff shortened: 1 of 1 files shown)


ladybug-me:
Yes, having a single generator function instead would be much better/bug-free for future updates.

ladybug-me (changes requested):
Please update this PR to use a single, shared generator function for both window list builders just as you suggested

WinTone01:
Done — pushed as a separate commit on top so the original fix stays readable.

Both producers now interpolate one `kWindowListJs` block: the entry shape, the desktop-index lookup and the skip rules live there, and the persistent bridge script and the one-shot refresh each add only their own `callDBus`. Nothing describes a window entry twice any more.

Verified the refactor is behaviour-preserving rather than assuming it: ran the shared generator against the live KWin scripting API and compared its output to what the running bridge was publishing at the same moment — same window count, same field set, no differing values. Also syntax-checked the generated scripts, both the shared block alone and the concatenation that gets injected.

ladybug-me (approved):

PR #427, opened 2026-08-06
feat: add dodge windows mode to the bar

### LINKED ISSUE #406: Dodge windows for taskbar
Window geometry is stable enough now, implement bar autohide on window intersection

WinTone01:
## What does this change?

Closes #406.

A persistent bar always sits on screen and reserves an exclusive zone for itself. Dodge mode keeps it visible over empty desktop but retracts it as soon as a window covers the strip it occupies — the behaviour Plasma panels call "Dodge Windows".

Overlap is evaluated per screen against the window geometry the KWin bridge already publishes, skipping minimized windows and windows on other virtual desktops, so a window you cannot see does not push the bar away. The hover edge and the usual toggles still reveal the bar while it is dodging, so nothing becomes unreachable.

Exposed as a toggle under Taskbar → Behaviour, disabled by default, and greyed out when the bar is not persistent (a non-persistent bar is already hidden, so the mode has nothing to add).

## How did you test it?

- On a live shell: enabled the toggle, moved a window over the bar and watched it retract, moved it away and watched it come back, and confirmed the hover edge still reveals it mid-dodge.
- Checked a maximized window against a real output: KWin insets maximized windows by the reserved struts, so they measure 2540x1370 on a 2560x1440 screen and correctly do not count as covering it.
- Left it off (the default) and confirmed the bar behaves exactly as before.

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

The one design decision worth flagging: the bar stops reserving an exclusive zone while this mode is on. That is load bearing rather than incidental. With a zone reserved, windows are kept off the bar, so nothing ever overlaps it, the bar comes back, the zone is reserved again — it oscillates every frame. Letting windows sit under the bar is what makes the overlap test mean anything.

Consequence to be aware of: with dodge on, a maximized window covers the full output rather than stopping at the bar, so the bar dodges for it. That matches how Plasma's own dodge mode behaves, but it is a visible difference from the default and is why this ships off.

### DIFF
--- a/shell/modules/background/Background.qml
+++ b/shell/modules/background/Background.qml
@@ -84,7 +84,7 @@ Variants {
         Loader {
             id: clockLoader
 
-            readonly property int clockBarZone: Visibilities.bars.get(win.modelData.name)?.exclusiveZone ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
+            readonly property int clockBarZone: Visibilities.bars.get(win.modelData.name)?.visualThickness ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
             readonly property int clockBaseMargin: Tokens.padding.extraLargeIncreased
 
             asynchronous: true
@@ -200,7 +200,7 @@ Variants {
         Loader {
             id: lyricsLoader
 
-            readonly property int lyricsBarZone: Visibilities.bars.get(win.modelData.name)?.exclusiveZone ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
+            readonly property int lyricsBarZone: Visibilities.bars.get(win.modelData.name)?.visualThickness ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
             readonly property int lyricsBaseMargin: Tokens.padding.large * 2
 
             asynchronous: true
--- a/shell/modules/background/DesktopIcons.qml
+++ b/shell/modules/background/DesktopIcons.qml
@@ -108,7 +108,7 @@ Item {
 
         anchors.fill: parent
         
-        readonly property int barZone: Visibilities.bars.get(root.screenData.name)?.exclusiveZone ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
+        readonly property int barZone: Visibilities.bars.get(root.screenData.name)?.visualThickness ?? (Tokens.sizes.bar.innerWidth + Math.max(Tokens.padding.small, Config.border.thickness))
 
         readonly property int baseMargin: Tokens.padding.large * 2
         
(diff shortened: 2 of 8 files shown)


ladybug-me:
As mentioned, Bar does not reveals itself when focused window is not intercepting, i.e., it hides if any opened window intercepts it.

Since, this pr is already focused on creating this feature, it will be great if you could add **Dodge focused windows only** feature too.

Its totally upto you if you want to add this feature.
Otherwise this pr is good to merge. (Verified)

Just say the word and I will merge it.

WinTone01:
Added it — thanks for verifying the base first.

`bar.dodgeFocusedOnly`, off by default, sits next to the existing toggle and is only selectable once dodging is on (on its own it decides nothing). With it enabled the bar only gets out of the way of the window actually in use, so a background window parked over the bar no longer keeps it retracted — which is the case you described, and it does read as the bar simply refusing to come back.

One small decision worth flagging: when nothing is focused, this dodges nothing rather than everything. That keeps the bar visible on an empty desktop, which seemed the more useful default of the two.

Pushed as a separate commit so the part you already verified stays reviewable on its own. Good to merge from my side whenever you are happy with it.

ladybug-me:
Verified, Nice work.

#425 was required for this to work.

**Bug**: Since exclusion zones drop to 0 when this feature is enabled, it causes the Quickshell Desktop objects to go behind the bar or should I say don't align according to the bar position. (Desktop icons, clock, etc.).

Easy visual way to reproduce: Bar on top, Desktop icons enabled, toggle Dodge windows in nexus.

Your decision to keep it visible on empty desktop is same as what I would I have implemented. Hence, this bug needs to be addressed.

#### The addressed feature is perfectly stable and can be merged successfully if you want.

WinTone01:
Fixed - `exclusiveZone` was doing two unrelated jobs.

It's the value reported to KWin for window-placement negotiation, and it's what `DesktopIcons`, the desktop clock, the desktop lyrics widget, the audio visualiser, and Shimeji's floor all read to know how much room the bar visually takes up. Dodge mode needs the first one to drop to near nothing while it's on - a reserved zone keeps windows off the bar, so nothing would ever overlap it and the mode could never engage - but that same drop was leaking into the second job, which has nothing to do with window-manager negotiation. With dodge on, all five of those consumers thought the bar took up only `Config.border.thickness` instead of its real content width, so they rendered under where the bar actually sits whenever it's visible.

Added `BarWrapper.visualThickness`: `exclusiveZone` without the dodge carve-out - what `exclusiveZone` always was before this feature existed. The five desktop-layer consumers now read that instead. `Exclusions.qml`, the one genuine window-manager consumer, is untouched.

## Testing

Reproduced your exact repro (bar on top, desktop icons on, dodge windows toggled on) and confirmed numerically via a temporary debug log: with dodging enabled, `exclusiveZone` correctly dropped to 10 (border thickness) while `visualThickness` stayed at 60 (the bar's real content width), and `DesktopIcons`'s margin followed `visualThickness`. Re-verified the non-dodge case is unaffected (both properties equal there, matching prior behaviour) and that dodge mode itself still engages correctly.

ladybug-me:
Very nice work.
Verified and works perfectly
Merging now

PR #428, opened 2026-08-06
feat: anchor minimize animations to the dock

### LINKED ISSUE #355: Link magic lamp animation to the dock instead of cursor
### What's the idea?

Right now, the magic lamp minimize/maximize animations are linked to the cursor. When the user moves the cursor during the animation, the window gets squashed, and it's not very pleasant to watch (I would attach a video but I don't want to break anything while removing/hiding the dock and adding it back in).

I fixed it by adding a KDE dock under the Caelestia one, but it's not very practical. Even though the KDE dock is visually under the Caelestia dock, it still physically takes up space. Because of this, every app going full-screen has a gap between the dock and the app:

Image

### Why would it be useful?

Just more pleasant to watch

WinTone01:
## What does this change?

Closes #355.

KWin's Magic Lamp animates a window into its icon geometry — the rect its taskbar entry occupies. Nothing was publishing one: the dock is a Quickshell surface rather than a Plasma panel, so every window had an invalid icon geometry and the effect fell back to guessing from the pointer. That is why the animation chased the cursor, and why putting a real Plasma panel under the dock worked around it, at the cost of that panel still reserving space.

Each dock tile's rect is now published through the taskbar side of plasma-window-management, the same mechanism Plasma's own Task Manager uses.

Cost of the change is small:

- The protocol XML is vendored next to the two already in `shell/plugin/src/Caelestia/protocols`, so this adds **no** runtime or packaging dependency.
- The shell's desktop file already requests `org_kde_plasma_window_management`, so it adds **no** new privilege.
- `Qt::GuiPrivate` comes along for `QPlatformNativeInterface`, which is the only way to reach a `QWindow`'s `wl_surface` — Qt exposes no public native interface for it. It ships with `qt6-base` on Arch and `qt6-qtbase-private-devel` on Fedora, both of which setup already installs.

## How did you test it?

Measured rather than eyeballed, since "the animation looks right" is hard to argue about:

- Queried `iconGeometry` for every window through the KWin scripting API. Before: `0,0 0x0` on every window, i.e. invalid, which is exactly the condition that makes Magic Lamp fall back to the cursor. After: each window reports a distinct real rect on the bar strip (`1066,1394 32x32`, `978,1394 32x32`, ...), matching its own dock tile.
- Confirmed visually that minimize now animates into the window's own tile, including when the pointer is parked over a different tile — that is the case where the old behaviour was obvious.
- Exercised the protocol in an isolated second Quickshell instance before letting it near a real session, checking that the extension binds, that a valid and a bogus uuid both behave, and that the process exits cleanly with no coredump.
- Running in a live shell since.

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

Three things the protocol forced, each with a reason that is not obvious from the diff:

**Proxies are released from `aboutToQuit`, not from destructors.** Releasing a Wayland proxy is itself a request, so it has to happen while the connection is up. A QML singleton's children are destroyed from an exit handler, long after Qt has closed the display, and marshalling there segfaults the shell as it quits. A latched flag makes anything destroyed later leak its proxy rather than touch it. Same reason the management global is owned by the singleton instead of living in a function-local static. (#426 fixes a pre-existing instance of this same trap in the screencast code; the two are independent.)

**`setGeometry` takes the tile item, not a window and a rect.** The protocol wants surface-relative coordinates, which is exactly what mapping the item into its own scene gives — and QML cannot pass a `QWindow` to C++ at all, it throws before the call is made.

**Rects are re-read on a slow timer rather than from bindings.** A tile moves for reasons no single binding can watch: the bar sliding in and out, the dock list scrolling, a drag reordering tiles. `mapToScene` is a plain call that never re-runs by itself. Unchanged rects are dropped before they reach the wire, so a dock at rest sends nothing; if you would rather see this driven by explicit change signals I am happy to try, but I could not find a set that covers all three cases.

This touches `Dock.qml`, as does #424 — different regions, and merging both locally was conflict-free, but flagging it in case order matters to you.

### DIFF
--- a/shell/modules/bar/components/Dock.qml
+++ b/shell/modules/bar/components/Dock.qml
@@ -36,6 +36,20 @@ Item {
 
     HoverHandler { id: dockHover }
 
+    // Drives the icon-geometry publishing below. A tile's rect on screen moves
+    // for reasons no single binding can watch — the bar sliding in and out, the
+    // dock list scrolling, a drag reordering tiles — and mapToItem() is a plain
+    // call that never re-runs on its own. Re-reading a handful of rects twice a
+    // second covers all of it, and MinimizeGeometry drops rects that have not
+    // actually changed, so a steady dock sends nothing over the wire.
+    Timer {
+        id: minimizeGeometryTicker
+
+        interval: 500
+        repeat: true
+        running: root.visible
+    }
+
     ListModel { id: dockModel }
 
     function saveNewOrder(): void {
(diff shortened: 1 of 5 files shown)


ladybug-me:
Please see #437

PR #429, opened 2026-08-06
feat: resolve icons for Wayland windows through the compositor

WinTone01:
## What does this change?

The extractor reads `_NET_WM_ICON` straight off the window, which only exists for XWayland clients. A native Wayland window with no resolvable desktop entry has no X window to read and no themed icon to look up, so the dock drew a generic placeholder for it.

Games are the case that matters. Launched through Proton's Wayland path a game reports its own exe name as the app id — `watchdogs2.exe` — which matches no desktop entry, and there is no X window behind it either, so both existing sources come up empty.

KWin does know the icon: it resolves it from the app id, from `xdg_toplevel_icon_v1`, or from whatever the client set, and hands it to taskbars over plasma-window-management. This asks for it there when the X read finds nothing — the same source Plasma's own Task Manager draws from on Wayland.

Icons land in the same content-addressed cache the X path writes to, so when both sources produce the same icon they share one file.

## How did you test it?

Against a real session that had gone fully Wayland (`_NET_CLIENT_LIST` was empty — not a single X client left, so the existing extractor could do nothing for anything on screen):

- Asked for the icon of every live window through an isolated second Quickshell instance. Five of seven resolved, including `watchdogs2.exe` and `explorer.exe`, which previously had nothing.
- The strongest check: the same game had been running under XWayland earlier in the day, and the X extractor had produced `ee9ba6a0fb3988ac.png` for it. Under Wayland, the compositor path resolved to **the same file** — the cache is content-addressed, so that is a byte-for-byte match confirming the icon is genuinely the game's and not something adjacent.
- Ran the whole chain through the QML service (`WinIcons.request` → `sourceFor`) for a Wayland window and got the cached PNG back where it would previously have returned a themed-icon fallback.
- Checked the process exits cleanly with no coredump, since this adds proxies with the lifetime constraints described in #428.
- Installed and running in a live shell.

Two windows did not resolve (`com.anthropic.Claude`, `heroic`). Both have desktop entries, so the dock already draws them correctly and never reaches this path — but worth stating that a null answer from the compositor is a normal outcome, not an error, and simply leaves the existing fallback in place.

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

The compositor writes the icon into a pipe from its own thread, so it cannot be read inline — the write blocks until someone drains it. The read is non-blocking on a `QSocketNotifier` and the result arrives on a signal; repeat asks while one is in flight are dropped.

The `org_kde_plasma_window` handle bookkeeping moves out of `MinimizeGeometry` into a shared owner, because two services now want the same handle for the same window and opening two of them would be wasteful. That owner is deliberately leaked rather than held in a destructible static, and releases its proxies from `aboutToQuit` — same constraint as #428, and the same trap #426 fixes in the screencast code.

One thing I deliberately did not do: subscribe to `themed_icon_name_changed` / `icon_changed` to keep icons live. Window icons effectively never change after mapping, and a one-shot ask keeps this much simpler. Easy to add if you would rather it track.

### DIFF
--- a/shell/modules/bar/components/Dock.qml
+++ b/shell/modules/bar/components/Dock.qml
@@ -697,7 +697,7 @@ Item {
                 // these (every unmapped Proton title is "steam_app_default").
                 const pid = ipc.pid || 0;
                 if (!entry)
-                    WinIcons.request(appClass, ipc.title || "", pid);
+                    WinIcons.request(appClass, ipc.title || "", pid, ipc.address ? String(ipc.address) : "");
 
                 apps.push({
                     id: appClass,
--- a/shell/modules/launcher/items/WindowSwitcherItem.qml
+++ b/shell/modules/launcher/items/WindowSwitcherItem.qml
@@ -44,7 +44,7 @@ Item {
         opacity = 1;
 
         if (root.modelData) {
-            WinIcons.request(root.modelData.class, root.modelData.title, root.modelData.pid ?? 0);
+            WinIcons.request(root.modelData.class, root.modelData.title, root.modelData.pid ?? 0, root.modelData.address ? String(root.modelData.address) : "");
         }
 
         Qt.callLater(() => {
--- a/shell/plugin/src/Caelestia/Services/CMakeLists.txt
+++ b/shell/plugin/src/Caelestia/Services/CMakeLists.txt
@@ -59,6 +59,10 @@ qml_module(caelestia-services
         windowicon.hpp
         minimizegeometry.cpp
         minimizegeometry.hpp
+        plasmawindows.cpp
+        plasmawindows.hpp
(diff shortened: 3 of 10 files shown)


ladybug-me (changes requested):
Please rebase

WinTone01:
Rebased onto `dev`. Now that #424 and #428 have landed, the stacked commits are gone and this is a single commit touching 10 files — the note at the top of the description no longer applies, so I have removed it.

Rebuilt on the new base and re-ran the checks: builds clean, `check_config_references.py` passes, and the QML convention linter reports one violation fewer than `dev` does for the files this touches.

PR #438, opened 2026-08-06
fix: give every dock tile its own minimize rect

### LINKED ISSUE #437: Magic lamp failing for top, right, left bar positions
Only works for bottom position

**Hint**: Setting bar position to left or right explains the situation.

WinTone01:
## What does this change?

Closes #437.

Icon geometry was read from each delegate, and a delegate reports position 0 here — measured, across every index. So every window was published the same rect: whichever tile ran last won, and every app animated into that one spot. With the bar anywhere but the bottom that spot was not even on the bar, which is what made it look like only bottom worked.

The rects are now derived from the list's own geometry and the entry's index, which is orientation-agnostic and needs nothing from the delegates.

Two things fell out of measuring it:

**The component is instantiated where it has no bar to sit on** — with no window, or laid out to zero size. Those copies were publishing too and overwriting the real dock's rects; one was reporting a position off the bottom of the screen. They are skipped now.

**Tiles scrolled out of the list** were keeping whatever was published for them last. That is stale across a scroll and badly wrong across a bar move — a vertical bar fits about a third as many tiles, so the rest were still pointing at their old horizontal positions. They are clamped to the nearest edge of the list instead, which is where they would scroll back in from.

## How did you test it?

By asking KWin for the actual `iconGeometry` of every window rather than eyeballing the animation.

Bottom bar, after the fix — each window has its own rect, on the bar strip, stepping by one tile:

```
zen=1000,1394  discord=1044,1394  Claude=1088,1394  Ferdium=1220,1394  Yandex=1572,1394
```

Before, every one of those was the same `956,1442`.

Left bar — x pinned to the edge, y stepping instead, confirming the vertical path:

```
zen=10,699  discord=10,751  Claude=10,803
```

The shell's own view agreed with KWin's in both cases. I also confirmed a plain ListView with these transitions positions its delegates normally, so the delegates reporting 0 is specific to this dock rather than something to fix in the view.

Top and right go through the same two branches as bottom and left with a different origin, so I did not measure them separately.

**Not measured:** the clamp for scrolled-out tiles. I found that problem in the left-bar measurement above (two windows still pointing at their old bottom-bar rects) and fixed it after, but did not re-measure with a vertical bar afterwards. The logic is a one-line clamp into the list's own span; flagging it rather than implying it was verified like the rest.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

I did not chase down *why* a delegate reads position 0 here — the isolated ListView test says it is not the view's normal behaviour, so something in this dock's delegate setup is doing it. Deriving from the list geometry sidesteps it entirely and is arguably the more robust source anyway, but if you would rather have the underlying cause found I am happy to keep digging.

### DIFF
--- a/shell/modules/bar/components/Dock.qml
+++ b/shell/modules/bar/components/Dock.qml
@@ -88,13 +88,29 @@ Item {
             // The nearest edge is where it would scroll back in from.
             const offset = Math.max(0, Math.min(i * step - scrolled, span - size));
 
-            const x = Math.round(horizontal ? origin.x + offset : origin.x);
-            const y = Math.round(horizontal ? origin.y : origin.y + offset);
+            let x = Math.round(horizontal ? origin.x + offset : origin.x);
+            let y = Math.round(horizontal ? origin.y : origin.y + offset);
+            let w = size;
+            let h = size;
+
+            if (Config.bar.position === "left") {
+                w = x + size;
+                x = 0;
+            } else if (Config.bar.position === "right") {
+                const winW = win ? win.width : 0;
+                w = winW > 0 ? winW - x : size;
+            } else if (Config.bar.position === "top") {
+                h = y + size;
+                y = 0;
+            } else if (Config.bar.position === "bottom") {
+                const winH = win ? win.height : 0;
+                h = winH > 0 ? winH - y : size;
+            }
 
             for (let j = 0; j < tops.length; j++) {
                 const address = tops[j]?.address;
                 if (address)
-                    MinimizeGeometry.setGeometry(root, String(address), x, y, size, size);
+                    MinimizeGeometry.setGeometry(root, String(address), x, y, w, h);
             }
         }
     }
--- a/shell/modules/bar/components/Dock.qml
+++ b/shell/modules/bar/components/Dock.qml
@@ -36,21 +36,68 @@ Item {
 
     HoverHandler { id: dockHover }
 
-    // Drives the icon-geometry publishing below. A tile's rect on screen moves
-    // for reasons no single binding can watch — the bar sliding in and out, the
+    // Re-publishes the icon geometry below. A tile's rect on screen moves for
+    // reasons no single binding can watch — the bar sliding in and out, the
     // dock list scrolling, a drag reordering tiles — and mapToItem() is a plain
     // call that never re-runs on its own. Re-reading a handful of rects twice a
     // second covers all of it, and MinimizeGeometry drops rects that have not
     // actually changed, so a steady dock sends nothing over the wire.
     Timer {
-        id: minimizeGeometryTicker
-
         interval: 500
         repeat: true
         running: root.visible
+        onTriggered: root.publishMinimizeGeometry()
     }
 
     ListModel { id: dockModel }
+
+    // Tell KWin where each app's taskbar entry sits, so minimize/restore
+    // effects animate into the dock instead of guessing from the cursor.
+    //
+    // The rects are derived from the list's own geometry and the entry's index
+    // rather than from each delegate: a delegate reports position 0 here, so
+    // asking it gave every window the same rect — whichever tile published last
+    // won, and every app animated to that one spot.
+    //
+    // This component is also instantiated where it has no bar to sit on (with
+    // no window, or laid out to zero size). Those copies have no meaningful
+    // rect to offer and would otherwise overwrite the real dock's, so they are
+    // skipped rather than allowed to publish nonsense.
+    function publishMinimizeGeometry(): void {
+        const win = QsWindow.window;
+        if (!win || listView.width <= 0 || listView.height <= 0)
+            return;
+
+        const horizontal = bar.isHorizontal;
+        const origin = listView.mapToItem(null, 0, 0);
+        const size = Math.round(container.itemSize);
+        const step = container.itemSize + root.spacing;
+        const span = horizontal ? listView.width : listView.height;
+        const scrolled = horizontal ? listView.contentX : listView.contentY;
+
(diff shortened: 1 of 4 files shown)


ladybug-me:
The actual issue still exists
Please refer these two videos.
**Note:** In both references, see how the animation collapses from the window edge corresponding to the destination (dock icon). Bar position: Left

**Bug**

https://github.com/user-attachments/assets/a6e47a7c-586c-4581-bcd8-25fcc7364a90


**Required output**

https://github.com/user-attachments/assets/343e1cbe-fba0-40a8-a68d-9f75ea1ff3dc


Required output is from main branch and hence, I did not click the minimize button as that would animate to minimize button itself. But its the same when clicked on dock icon to minimize.

Another thing is that the animation delay is still there before restoring the window hence proving the magic lamp plugin is not crashing.

WinTone01:
Found the actual root cause - it's not in the per-tile geometry math from the previous commit, that part was correct. It's that KWin's own Magic Lamp effect never had a working way to use it, on any bar position.

## What's actually happening

Magic Lamp's minimize animation (`effects/magiclamp` in KWin) looks up which panel owns the icon it's animating to by walking the window stacking order for a window where `isDock()` is true and whose geometry intersects the published icon rect:

```cpp
for (EffectWindow *window : stackingOrder) {
    if (!window->isDock()) continue;
    if (window->frameGeometry().intersects(icon)) { panel = window; break; }
}
```

Our drawers surface - the one window that actually renders the bar/dock, spanning the whole screen - was never reporting as a dock, so this never found it, on any orientation.

`StyledWindow.qml` hardcodes `WlrLayershell.namespace` to `"panel"` (or `"desktop"` for `isDesktopWidget`). The `name` property every `StyledWindow` declares is a local label with no effect on the Wayland namespace at all, despite reading like it should be one. KWin's `LayerShellV1Window::scopeToType()` maps the namespace `"dock"` to `WindowType::Dock` and anything not in its fixed list - `"panel"` included - to `WindowType::Normal`. Confirmed directly against KWin's own scripting API: before this fix, `workspace.windowList()` reported `dock=false` for the drawers window on both of my monitors; after it, `dock=true` on both.

## Why this explains both bugs

With no panel found, Magic Lamp falls through to its "no panel" heuristic: check whether the icon rect touches a screen edge by exact pixel equality, defaulting to Bottom if none do:

```cpp
if (rect.left() == iconScreen.left()) position = Left;
else if (rect.right() == iconScreen.right()) position = Right;
else if (rect.top() == iconScreen.top()) position = Top;
else position = Bottom;
```

Our published rects sit a few pixels in from the true edge (padding), so that check never passes - every orientation silently got Bottom's warp math regardless of where the bar actually was. That's:

- **The wrong-direction animation in the video on this PR** - Bottom's math used for a Left bar produces a warp shape that doesn't match a vertical bar at all.
- **This session's actual report** (top never animates on either the dock icon or the window's own minimize button) - Bottom's math assumes the icon sits below the window; for a Top bar it's the opposite, and the math degenerates to nothing visible instead of just looking wrong.

## The fix

Sets `WlrLayershell.namespace: "dock"` directly on the drawers window, since that's the one surface Magic Lamp needs to find. The shared `StyledWindow` default is untouched - every other panel-style window in the shell (OSD, notifications, tooltips, etc.) keeps reporting as a plain panel, not a dock.

## Testing

Confirmed `dock=true` via KWin's scripting API on both monitors after the change. Reproduced the exact reported bug (Top: no animation on either trigger) and confirmed it now plays correctly, aimed at the dock. Re-verified Bottom, Left and Right still animate correctly - no regression there - and that general shell interactivity (dashboard, sidebar, launcher, window switching / Alt+Tab exclusion) is unaffected by the window now reporting as a dock.

ladybug-me:
Great, the top and bottom now work perfectly.

**The left and right still have the same bug.**

Reason: Instead of left | right collapse animation for left | right bar position, it still plays bottom collapse animation, now sometimes top collapse animation too, as demonstrated in bug video.

Why it seems to work: The window title bar is simply above the dock, as in the bug video.

WinTone01:
> Great, the top and bottom now work perfectly.
> 
> **The left and right still have the same bug.**
> 
> Reason: Instead of left | right collapse animation for left | right bar position, it still plays bottom collapse animation, now sometimes top collapse animation too, as demonstrated in bug video.
> 
> Why it seems to work: The window title bar is simply above the dock, as in the bug video.


https://github.com/user-attachments/assets/bf0601e6-56bd-4b4f-af06-a889527c7ed4

WinTone01:
I don't understand what do you mean?  I think there is no bug can you check the video @ladybug-me

ladybug-me:
yes as you can see its still playing the bottom collapse animation causing the following bug:

https://github.com/user-attachments/assets/b8b678f2-70ac-41c6-aedf-009bf1919c9a

Look how it should be, to be able to work (When I turn it to left position):

https://github.com/user-attachments/assets/fffabc0a-aed9-4afb-bae4-8278ff84fbf0

WinTone01:
This is weird when i try this with other application it's works but if i try with foot this is not work can you explain me why?  
*my brain is burning :laughing:

ladybug-me:
Found a fix.

Instead of setting the namespace as dock, we set it to panel and stretch the dock geometry by 8px to touch the edge of monitor

Will push the commit in a sec

WinTone01:
https://github.com/user-attachments/assets/9257ef42-c601-4608-903f-05bf2f82b18b

ladybug-me:
Ok try now

PR #441, opened 2026-08-07
Debian support

### LINKED ISSUE #200: [Feature] Debian Support
### Proposal

Add support for Debian-based distributions, especially Kubuntu, for this KDE theme/setup.

### Use case

I previously used this theme on CachyOS KDE and it worked great. I switched to Kubuntu for stability reasons and would like to continue using this setup there as well. Debian-family support would make it possible to keep the same workflow after switching distros.

### Additional context

Could this be considered for a future update? Thanks for your work on this project.

ladybug-me:

## Type

- [ ] Bug fix
- [x] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

### DIFF
--- a/installer/src/UI.cpp
+++ b/installer/src/UI.cpp
@@ -285,10 +285,10 @@ while (!g_quit) {
     }
 
     string distro_select() {
-        vector<string> options = {"Arch-based", "Fedora", "Exit"};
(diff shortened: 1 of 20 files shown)


ladybug-me:
Tested on Kubuntu 26.04

PR #444, opened 2026-08-08
feat: take screen corners from KWin while the overview wants them

### LINKED ISSUE #443: Steal KDE hot corner ( Screen Edges )
Caelestia overview must not interfere with KDE overview or any other corner effects set in KDE settings.

Just like stealing KWin shortcuts and restoring them on exit, we must find a way to steal the corner from KWin on the go.

If not possible, persistent writing to disk (Set corner to none) will also do.

**Requirements**: Toggling a corner for Overview in nexus must immediately take effect in live session, hence if a adjacent corner was set for KDE effects, they shouldn't interfere with Caelestia overview.

WinTone01:
Closes #443.

## The D-Bus part

There isn't a signal — KWin doesn't subscribe to `org.kde.kconfig.notify` for `kwinrc`, which is why no amount of `kwriteconfig6 --notify` did anything. It re-reads only when something calls the method:

```bash
qdbus6 org.kde.KWin /KWin org.kde.KWin.reconfigure
```

That covers effect-registered edges too, not just the built-in ones — `OverviewEffect::reconfigure()` does `OverviewConfig::self()->read()` and then unreserves/re-reserves its borders, so one call is enough for everything.

Handy for checking it worked, since it reads back live state rather than config:

```bash
qdbus6 org.kde.KWin /KWin org.kde.KWin.supportInformation | sed -n '/Screen Edges/,/^$/p'
```

## Why it still wouldn't have worked

`[ElectricBorders]` is only one of the things that can hold a corner. Any `[Effect-*]`, `[Script-*]` or `[TabBox]` group can reserve an edge through a `BorderActivate` int-list, and KDE's Overview effect defaults to `[Effect-overview] BorderActivate=7` — top-left — per `overviewconfig.kcfg`. On a stock install that key isn't in `kwinrc` at all, so the corner KDE actually fires from was never being touched. That's the conflict in the issue.

Third thing: turning a corner back off wrote the literal `"Overview"`, which overwrites whatever the user had there rather than restoring it.

## What's in here

A `ScreenEdges` service that claims a corner by neutralising *every* claimant of that edge:

- `[ElectricBorders] <Corner>` → `None`
- every `BorderActivate`-shaped key in any `Effect-*` / `Script-*` / `TabBox` group, with only that edge removed from the list — unrelated edges in the same list survive
- keys KWin reserves by compiled-in default with nothing in `kwinrc`, written out explicitly (otherwise KConfig just falls back to the default we're trying to displace)

Originals are stashed verbatim in `~/.config/caelestia/stolen-screen-edges.json`, including a null meaning "this key was absent, delete it on restore", and put back on release, on clean exit, and on next startup after a crash — same shape as the KWin global shortcuts in `globalshortcut.cpp`. `ScreenCorners.qml` binds it to the four overview settings, so corners are claimed at startup too, not only when toggled in Nexus.

## Tested live

Set up a three-way conflict (`Effect-overview=7`, `ElectricBorders/TopLeft=ShowDesktop`, `TabBox=7,2`) with Caelestia wanting top-left:

| | result |
|---|---|
| claim | `Effect-overview` 7→9, `TopLeft` ShowDesktop→None, `TabBox` `7,2`→`2`, `actionTopLeft` 1→0 live |
| clean exit | all three back byte-for-byte, `actionTopLeft` 0→1 live |
| `kill -9` then restart | stash survived, originals restored, then re-stolen cleanly |
| stock install (key absent) | explicit `9` written, stash records null, exit removes the key and leaves no trace |

`kwinrc` came out identical to its pre-test backup afterwards.

One thing worth your call: I left touch borders (`TouchBorderActivate*`) alone, on the grounds that a corner isn't really touch-activatable. Easy to fold in if you'd rather.

### DIFF
--- a/shell/modules/ScreenCorners.qml
+++ b/shell/modules/ScreenCorners.qml
@@ -1,0 +1,38 @@
+import QtQuick
+import Quickshell
+import Caelestia.Config
+import Caelestia.Services
+
+// Keeps KWin's hold on the four screen corners in sync with which ones the
+// overview wants. Bindings, not one-shot writes on toggle, so a corner is
+// taken at startup too and given back the moment it stops being wanted —
+// ScreenEdges itself handles putting KWin's config back exactly as it found
+// it, including after a crash.
+Scope {
+    id: root
+
+    readonly property bool overviewOn: Config.overview.enabled
+    readonly property bool wantTopLeft: overviewOn && Config.overview.hoverTopLeft
+    readonly property bool wantTopRight: overviewOn && Config.overview.hoverTopRight
+    readonly property bool wantBottomLeft: overviewOn && Config.overview.hoverBottomLeft
+    readonly property bool wantBottomRight: overviewOn && Config.overview.hoverBottomRight
+
+    function sync(corner: int, wanted: bool): void {
+        if (wanted)
+            ScreenEdges.claim(corner);
+        else
+            ScreenEdges.release(corner);
+    }
+
+    onWantTopLeftChanged: sync(ScreenEdges.TopLeft, wantTopLeft)
+    onWantTopRightChanged: sync(ScreenEdges.TopRight, wantTopRight)
+    onWantBottomLeftChanged: sync(ScreenEdges.BottomLeft, wantBottomLeft)
+    onWantBottomRightChanged: sync(ScreenEdges.BottomRight, wantBottomRight)
+
+    Component.onCompleted: {
+        sync(ScreenEdges.TopLeft, wantTopLeft);
+        sync(ScreenEdges.TopRight, wantTopRight);
+        sync(ScreenEdges.BottomLeft, wantBottomLeft);
+        sync(ScreenEdges.BottomRight, wantBottomRight);
+    }
+}
--- a/shell/modules/nexus/pages/panels/OverviewPanel.qml
+++ b/shell/modules/nexus/pages/panels/OverviewPanel.qml
@@ -140,41 +140,32 @@ PageBase {
             visible: !GlobalConfig.overview.showOnHover
         }
         SectionHeader {
-            text: qsTr("Please disable KDE's adjacent Screen corner settings to avoid conflicts")
+            text: qsTr("Corners")
         }
+        // Enabling a corner takes it off KWin for as long as it stays enabled;
+        // whatever KDE had bound there comes back untouched when it's turned
+        // off, on shell exit, or on the next start after a crash.
         ToggleRow {
             first: true
             text: qsTr("Top-Left corner")
             checked: GlobalConfig.overview.hoverTopLeft
-            onToggled: {
-                GlobalConfig.overview.hoverTopLeft = checked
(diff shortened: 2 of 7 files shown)


WinTone01:
Pushed two follow-ups.

**Touch borders.** An effect reserves pointer and touch edges through separate keys, so neutralising only `BorderActivate` left `TouchBorderActivate`, `GridTouchBorderActivate` and the `TabBox` touch variants still pointing at a corner we'd taken. Now covered. The separate `[TouchEdges]` group still needs nothing — it only has `Top`/`Right`/`Bottom`/`Left`, so a corner can't be bound there at all.

**The double-open.** Testing on a corner where `kwinrc` said `None` everywhere, KDE's overview still opened alongside ours. `reconfigure()` reloads the built-in edge actions but leaves an effect sitting on an edge it had already reserved — so KWin goes on firing it from a corner its own config says is free. `Effect::reconfigure()` is what unreserves and re-reserves from freshly read config, and per-effect that's:

```bash
qdbus6 org.kde.KWin /Effects org.kde.kwin.Effects.reconfigureEffect overview
```

So the global call is now followed by a `reconfigureEffect` on every effect with an edge key in `kwinrc` — whether or not we changed its value, since a stale reservation has to be flushed even when what's on disk was already harmless. This also matters on release: without it a corner could stay stuck to an effect after we'd given it back.

Isolated it by reserving top-left in the live overview effect, setting the config back to `None` behind its back, then confirming a global `reconfigure` left it firing while the per-effect call cleared it. Polling `org.kde.kwin.Effects.activeEffects` across a corner-hover test: `overview` and `screenedge` before, neither after.

ladybug-me:
**Results**: KDE Settings are changed correctly .

**But changes are not visible** in live session instantly which is the behavior we require.

My setup:
`Top-Left: Overview`
`Top-Right: Show Desktop`
`Bottom-Right: Grid`

After Stealing:
Overview and Grid not affected
**Show Desktop was stolen correctly**

After Re-login/reboot:
All the corners were correctly assigned to Caelestia

Same results when trying to restore them by Disabling the respective corner in nexus.

ladybug-me:
@WinTone01 
Let me test the new commit

ladybug-me:
@WinTone01 
Awesome Work!!
Verified, works perfectly..

Been messing with me for a week.. Thx a lot..

ladybug-me:
Just change the default overview corner to `top-left` and set `top-right` to false in `shell/plugin/src/Caelestia/Config/overviewconfig.hpp`

Then we are good to merge this one

WinTone01:
> @WinTone01 Awesome Work!! Verified, works perfectly..
> 
> Been messing with me for a week.. Thx a lot..

@ladybug-me Not problem dude :face_holding_back_tears:

WinTone01:
Done — `hoverTopLeft` now defaults to true and `hoverTopRight` to false, which also lines up with KWin's own default for the Overview effect (the corner we take from it).

PR #445, opened 2026-08-08
fix: stop overview cards overlapping, centre the grid, and cut the open cost

WinTone01:
Window cards were overlapping and running off the page, the grid arrived from a wide spread and converged inward on open, and the transition dropped frames both ways.

## Cards overlapping and overflowing

The layout hands out boxes that tile a page without overlapping, but the cards did not use them — each sized itself from its content:

```qml
implicitWidth:  cardLayout.implicitWidth + Tokens.padding.medium * 2
implicitHeight: cardLayout.implicitHeight + Tokens.padding.medium * 2
```

so every card came out a padding wider and a padding plus a caption taller than its own slot. Position right, size wrong. Cards are now exactly their box.

The grid also ran the full width of the screen, while opening the overview pulls the desktop into a rounded inset (`ContentWindow` grows the border to 15% of the short edge). The outer cards sat out on the black frame rather than the wallpaper. It now insets by that border. Worth noting for anyone touching this later: `panels`' own margins look like the right thing to use and are not — on whichever edge holds the bar, that margin is the bar's size instead.

## Two-row layouts not centred

Both packers place rows as they go and try to keep the block centred while doing it, which only holds when nothing gets clamped. Once a row is scaled down to fit the width, GNOME's running offsets leave a two-row layout flush against the bottom — measured at 179px of empty space above and zero below.

Rather than repair that arithmetic, both layouts now end by measuring the finished bounding box and centring that (`layoututils.hpp`). Exact whatever the packer did, one extra pass over a handful of rects. Verified on both layout modes: gaps went from `top=179 bottom=0` to `top=79 bottom=79`.

## Frame drops (unresolved)

Screencast streams are no longer requested for pages further than one swipe from the one in view — a workspace three pages away was being captured for nothing.

The grid also sized itself from the border thickness while that was still animating from 10px to its target, so the first layout landed on a nearly full-screen area and every card slid inward as it shrank. That is the cards-arriving-from-everywhere effect on open. Measured, the layout ran against 2540x1420 and a dozen intermediate sizes before settling at 2076x956; `ContentWindow` now also exposes the settled value, and the grid uses that, so it goes straight to the final size.

Also drops two `qDebug` lines from `calculateLayout` and two `console.log` calls from the page delegate, which ran per page per window-list update.

**On the frame drops I could not find the cause.** Turning screencasts off entirely does not make an open any cheaper, so creating them is not what costs. My per-open CPU benchmark swings by 3x run to run depending on what the captured windows are doing, so any figure from it is not worth quoting — I have dropped the one that was here. Left open for now.

## Design

Cards get a real surface and outline instead of floating thumbnails; the title moves off the floor and into the card next to its app icon; the active window is picked out in the accent colour; hovering lifts a card slightly, and the grid reserves exactly the room that lift needs so a card on the edge of the stage cannot grow off the wallpaper. Captions drop below 96px tall, where a title strip costs more than it tells you.

Tested live on KDE Wayland in both layout modes.

### DIFF
--- a/.github/workflows/test-dependencies.yml
+++ b/.github/workflows/test-dependencies.yml
@@ -36,7 +36,7 @@ jobs:
           # Give builduser ownership of the workspace so it can run the script
           chown -R builduser:builduser .
           su - builduser -c "cd $GITHUB_WORKSPACE && ./sdata/arch-dist/installDP.sh || true"
-          
+
           if [ -s /home/builduser/.cache/caelestia-kde/failed_packages.txt ]; then
             FAILED=$(cat /home/builduser/.cache/caelestia-kde/failed_packages.txt | tr '\n' ' ')
             echo "::error::[ARCH] Failed to install packages: $FAILED"
--- a/.github/workflows/update-submodules.yml
+++ b/.github/workflows/update-submodules.yml
@@ -35,15 +35,15 @@ jobs:
             echo "All submodules are fully up to date."
           else
             BRANCH_NAME="chore/update-dependencies-$(date +%s)"
-            
+
             git checkout -b "$BRANCH_NAME"
             git config user.name "github-actions[bot]"
             git config user.email "github-actions[bot]@users.noreply.github.com"
-            
+
             git add .
             git commit -m "chore: update submodules to latest upstream commits"
             git push origin "$BRANCH_NAME"
-            
+
             gh pr create \
               --base dev \
               --head "$BRANCH_NAME" \
--- a/.github/workflows/version-release.yml
+++ b/.github/workflows/version-release.yml
@@ -164,7 +164,7 @@ jobs:
           set -euo pipefail
 
           echo "Attempting to open a PR to sync main into dev..."
-          
+
           gh pr create \
             --base dev \
             --head main \
--- a/shell/modules/Shortcuts.qml
+++ b/shell/modules/Shortcuts.qml
@@ -55,7 +55,13 @@ Scope {
         description: "Toggle overview"
         onPressed: {
             const visibilities = Visibilities.getForActive();
-            visibilities.overview = !visibilities.overview;
+            // Pressing it again while the overview is up steps through the
+            // windows rather than dismissing it, so the same key that opened
+            // the grid also walks it — the thing every task switcher does.
+            if (visibilities.overview)
+                Visibilities.cycleOverview(false);
+            else
+                visibilities.overview = true;
         }
     }
     // qmllint disable unresolved-type
--- a/shell/modules/Shortcuts.qml
+++ b/shell/modules/Shortcuts.qml
@@ -55,13 +55,16 @@ Scope {
         description: "Toggle overview"
         onPressed: {
             const visibilities = Visibilities.getForActive();
-            // Pressing it again while the overview is up steps through the
-            // windows rather than dismissing it, so the same key that opened
-            // the grid also walks it — the thing every task switcher does.
-            if (visibilities.overview)
-                Visibilities.cycleOverview(false);
-            else
+            if (visibilities.overview) {
+                visibilities.overview = false;
+            } else {
+                if (typeof KWinActiveWindowBridge !== "undefined" && KWinActiveWindowBridge.activeWindow && KWinActiveWindowBridge.activeWindow.address) {
+                    Visibilities.preOverviewActiveWindowAddress = KWinActiveWindowBridge.activeWindow.address;
+                } else {
+                    Visibilities.preOverviewActiveWindowAddress = "";
+                }
                 visibilities.overview = true;
+            }
         }
     }
     // qmllint disable unresolved-type
--- a/shell/modules/background/Wallpaper.qml
+++ b/shell/modules/background/Wallpaper.qml
@@ -162,6 +162,10 @@ Item {
         readonly property bool isDynamicScheme: currentSchemeName.startsWith("dynamic")
         readonly property bool isDynamicMonochrome: isDynamicScheme && currentVariantName === "monochrome"
         readonly property bool needsMask: img.z === 1 && maskAnim.running
+        readonly property bool tiled: {
+            const m = Config.background.wallpaperFillMode;
+            return m === Image.Tile || m === Image.TileVertically || m === Image.TileHorizontally;
+        }
         readonly property bool shouldRecolor: Config.background.wallpaperRecolor
 
         function update(): void {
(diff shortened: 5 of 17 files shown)


ladybug-me:
I like the design but it requires some changes..

Let's chat on  [discord](https://discord.gg/9CX6EUaw)

WinTone01:
Thanks — pushed 1-6, and 7 I have to hand back to you unsolved.

**2, 3 — timer gone.** You were right that it did not belong. Testing your point 7 first: turning screencasts off entirely (`livePreviews: false`) does not make an open any cheaper, so creating the streams is not what costs, and the delay was buying nothing while putting the previews people came to click behind a timer. Streams are requested the moment the grid appears again. The one gate I kept is that pages further than one swipe from the one in view do not get streams at all — a workspace three pages away was being captured for nothing.

**4 — outline.** It marked whichever window was last focused. That reads as "already picked", exactly as you said. It now marks keyboard selection only, and the overview opens with nothing selected.

**6 — icon and title.** Same rule: they appear on the card being hovered or selected instead of on all of them at once, so a page of thumbnails is just thumbnails. Cards under 96px tall drop the caption entirely.

**5 — Meta+Tab.** The shortcut used to toggle, so the key that opened the grid closed it. Pressing it while the grid is up now steps the selection on. Tab and the arrow keys do the same, Return activates, Escape still closes.

**1 — GNOME.** That is roughly what 4 and 6 add up to: no chrome until a card is the one being looked at, selection as an outline rather than app state.

## 7 — could not isolate it

What I can say:

- Stream creation is not the dominant cost. With `livePreviews` off there are no streams at all and an open is no cheaper.
- Warming them does not help either — `continuousMode: true`, so streams are alive and reused rather than rebuilt, measured no better.
- The overview tears its whole item tree down on close (`Wrapper`'s `Loader` deactivates) and rebuilds it on every open. I tried keeping it loaded; that did not measurably help either.

The honest problem is that my measurement is not good enough to attribute this. I was sampling process CPU across the open, and the same build measures anywhere from 560ms to 2s depending on what the captured windows happen to be doing at the time — a playing video in one thumbnail moves it more than any change I made. I could not get `QSG_RENDER_TIMING` out of Quickshell's log filter to measure actual frame times instead, which is what this needs.

So I have also retracted the "1500ms -> 560ms" figure I put on the earlier commit and in this PR description. It came from that benchmark and it should not have been quoted.

If you have a way to get per-frame timings out of a running Quickshell I will pick this back up.

WinTone01:
Got frame timings out of it — `quickshell --log-rules "qt.scenegraph.time.*=true"` with `QSG_RENDER_TIMING=1`, then read the `.qslog` rather than stdout. That gives per-frame `sync` / `render` / `swap`.

## What the hitch actually is

One or two frames of **70-95ms**, and only on the **first** overview open of a shell session. Every open after that is clean — worst frame 16-18ms, i.e. plain vsync.

The whole thing is in `swap`:

```
en yavaş kare (toplam/sync/render/swap)
  82 / 1 / 7 / 73      <- first open
  16 / 0 / 0 / 16      <- every open after
```

`sync` and `render` are the shell's own CPU work and they never exceed 8ms. `swap` is the render thread blocked waiting for a buffer back from the compositor. So the shell is not doing too much work — it is being kept waiting.

## Streams are not the cause

Same cold open with `livePreviews: false`, so no screencast exists at all:

```
  81 / 1 / 4 / 75
```

Unchanged. Warming them the other way (`continuousMode: true`, streams alive and reused rather than rebuilt) does not help either. So requesting streams as the overview opens is not what drops the frames, and the delay I had added for it was buying nothing — that is gone.

## Also ruled out

Each of these measured on a cold first open, stall unchanged every time:

- **The window cards.** It happens on an empty workspace with zero cards.
- **KWin blur** (`appearance.blur: false`) and **overview blur** (`enableOverviewBlur: false`).
- **The full-screen `layer.enabled` FBO** on `overviewWallpaperLayer`.
- **First-time construction.** Pre-activating the overview tree and the second `Wallpaper` at startup so nothing is built on open: no change. Forcing an actual full-screen paint at startup (opacity 0.004 for 400ms) so buffers are allocated early: no change either.

And it is specific to the overview. Same cold-open measurement on the other drawers:

```
  dashboard   worst 31ms  (sync=25, swap=0)   <- CPU, first instantiation, expected
  launcher    worst 18ms  (swap=17)           <- vsync
  overview    worst 82ms  (swap=73)
```

## Where that leaves it

KWin's own CPU across the window is the same on the stalling cold open (260ms) and on a clean warm one (280ms) — it is not doing more work, and the shell is still blocked. That points at the buffer path rather than at anything the QML is doing, and I could not get further from this side.

If you have a way to see KWin's side of that first present — which surface it is waiting on and why the buffer is not released for ~5 vsyncs — I will pick it back up. Everything above reproduces with a single `caelestia shell -k` and one overview open.

WinTone01:
Looked at it from the compositor side too. I could not close it out, but the search space is a lot smaller now and one relevant fact turned up: this is **NVIDIA proprietary on Wayland** (RTX 3060, driver 610.57.04, KWin on EGL/OpenGL, atomic modesetting on).

## What the stall is made of

Two separate one-time costs, and I had been conflating them.

**a) Shader compilation — only after a rebuild.** One run showed `129ms total, render=58`. `render` is the shell's own GPU work, and 58ms of it is pipeline compilation. Qt caches compiled pipelines to disk, so this only bites on the first run after `~/.cache/qtshadercache-*` is invalidated, which is every time we rebuild. I confirmed it by deleting that cache: the compile then happens during shell *startup*, not on overview open, and the open goes back to `render=4`. So this is a developer-visible cost, not something users hit repeatedly. Worth knowing before anyone benchmarks a fresh build and blames the overview.

**b) The real one — ~70-95ms, entirely in `swap`, first overview open per shell session.** Reproduced on every configuration I tried:

```
83 / 0 / 4 / 78     first open   (total/sync/render/swap)
21 / 0 / 0 / 20     every open after
```

## Ruled out, each measured on a cold first open

- Screencast streams (`livePreviews: false` — no stream exists at all): unchanged.
- Warm streams (`continuousMode: true`): unchanged.
- The window cards: happens on an empty workspace with zero cards.
- KWin blur, overview blur, the full-screen `layer.enabled` FBO: unchanged.
- First-time construction: pre-activating the overview tree *and* the second `Wallpaper` at startup, so nothing is built on open — unchanged. Forcing a real full-screen paint at startup (opacity 0.004 for 400ms) to get buffers allocated early — also unchanged.
- Shader compilation, per (a).

And it is specific to this drawer. Cold opens of the others:

```
dashboard   worst 31ms  (sync=25, swap=0)   CPU, first instantiation, expected
launcher    worst 18ms  (swap=17)           vsync
overview    worst 83ms  (swap=78)
```

## Why I think it is below us

KWin's own CPU over the same window is 260ms on the stalling cold open and 280ms on a clean warm one — it is not doing more work, and the shell is still blocked ~5 vsyncs waiting for a buffer back. Nothing appears in KWin's journal at that moment. That combination points at the buffer/plane path on the first full-screen present rather than at anything the QML does, which fits NVIDIA on Wayland.

## The test I did not run

The obvious next one is whether KWin is trying to promote the now-full-screen shell surface to a hardware plane:

```bash
KWIN_DRM_NO_DIRECT_SCANOUT=1
```

That needs KWin restarted, which would take the session down, so I left it for you rather than doing it on a machine I am borrowing. If the stall disappears with it set, it is scanout/plane negotiation and there is likely nothing to fix on our side beyond not going full-screen-opaque in one step.

Reproduction is one `caelestia shell -k`, one overview open, with:

```bash
quickshell --log-rules "qt.scenegraph.time.*=true"
```

and `QSG_RENDER_TIMING=1`, then read the `.qslog` — the timings do not reach stdout.

WinTone01:
Found it, and 7 is fixed.

## What it was

The first time the shell paints the **whole screen**. Nothing else it draws is full-screen, which is exactly why no other drawer does this — cold, the dashboard's worst frame is 31ms and that is `sync` (CPU, first instantiation), the launcher never exceeds vsync, the overview blocks 70-95ms in `swap`.

`swap` is the render thread waiting for a buffer back. KWin burns the same CPU on the stalling open (260ms) as on a clean one (280ms), so nobody is doing more work — the shell is just kept waiting while the driver allocates for that first full-screen paint. NVIDIA proprietary on Wayland here (RTX 3060, 610.57.04), which is where that kind of first-allocation cost tends to show up.

## Fix

Paint the overview's wallpaper layer once, invisibly (`opacity: 0.004`, 400ms, 2.5s after start), so the allocation happens while nothing is animating. Cold shell, first open:

```
before   83ms worst frame (sync=0, render=4, swap=78), 1-2 frames dropped
after    19ms worst frame, nothing over 33ms
```

Verified there is no visible flash — screenshots across the warm-up window are indistinguishable from a normal desktop.

Worth noting: warming the **grid** does not work, only the wallpaper layer does. My first attempt warmed the grid and the open still stalled, because the grid is not what covers the screen. That is also why the earlier "pre-activate everything at startup" test came back negative — pre-building the tree is not enough, it has to actually paint.

## Ruled out getting there

Each measured on a cold first open, stall unchanged every time:

- **Screencast streams** — `livePreviews: false`, so no stream exists at all. Unchanged. Your hypothesis in 7 does not hold, and the delay I had added for it is gone.
- Warm streams (`continuousMode: true`).
- **The window cards** — happens on an empty workspace with zero cards.
- KWin blur, overview blur, the full-screen `layer.enabled` FBO.
- Constructing the tree — pre-activating the overview tree and the second `Wallpaper` at startup.
- **DRM / direct scanout.** I did not need to restart KWin for this: it reproduces identically under a nested `kwin_wayland --wayland-display`, which uses neither. So `KWIN_DRM_NO_DIRECT_SCANOUT=1` would not have told us anything.

## One thing for whoever benchmarks next

There is a second, separate cost that only appears on the first run after a rebuild: pipeline compilation, ~120-270ms of `render`. Qt caches compiled pipelines in `~/.cache/qtshadercache-*`, so it is invalidated every time we build and never hits users twice. I lost time to it before separating the two — deleting that cache moves the compile to shell startup and the open goes back to `render=4`.

Timings come from `quickshell --log-rules "qt.scenegraph.time.*=true"` with `QSG_RENDER_TIMING=1`, read out of the `.qslog` — they do not reach stdout.

PR #455, opened 2026-08-10
feat: replace KRunner with Vicinae, themed from the shell's palette

WinTone01:
Replaces KRunner with [Vicinae](https://github.com/vicinaehq/vicinae) and wires its colours to the shell's palette.

## KRunner keeps its key until the shell takes it

Vicinae goes on `Alt+Space` as an ordinary `CustomShortcut` in `Shortcuts.qml`. Nothing is disabled: the shell's `GlobalShortcut` already takes a key off whoever holds it when it binds one and gives it back on exit, so KRunner simply loses `Alt+Space` for as long as Caelestia is running. The shell keeps `Meta` and `Meta+Space` for its own launcher, so the two never fight over one.

Confirmed on a live session — with KRunner running, its `_launch` shortcut goes to no keys while the shell is up and comes back as `Alt+F2`, `Search` and `Alt+Space` when the shell quits, with the three recorded in `stolen-shortcuts.json` in between. A crashed session restores them on next start.

Two approaches were tried and dropped on the way. A `.desktop` entry under `[services]` in `kglobalshortcutsrc` never registers — `kglobalacceld` only reads those at login, and rebuilding ksycoca, restarting the daemon and clearing `NoDisplay` all left the component absent. And masking `plasma-krunner.service` turned out to be worse than redundant: it disables KRunner outright and is only undone by the uninstaller, so anyone who stopped running Caelestia without uninstalling would be left without it.

## Colours follow the wallpaper

The shell already publishes the scheme it generated to `~/.local/state/caelestia/scheme.json`, and Vicinae reads themes from its data directory and *watches* it. So the two only needed a translation: `shell/scripts/vicinae-theme.py` writes the Material palette out as a Vicinae theme, and a path unit reruns it whenever the scheme changes. Nothing restarts — the launcher repaints on its own.

Only the colours Vicinae cannot derive itself are written, so there is less to drift when either palette changes.

## Installing it

Upstream publishes no `.deb` or `.rpm`, only AppImages and a tarball. The tarball is laid out as a prefix (`bin/`, `libexec/`, `share/`, `lib/systemd/user/`) and resolves its helpers relative to the binary, so it unpacks into `/usr/local` — already on systemd's user unit search path, so the service it ships is found as-is. All three distros use it, so the only party trusted for the binary is the project itself.

What the AUR package was doing for us moves into the script:

- the download is **verified** against the digest upstream published, and refuses to install on a mismatch
- the input helper gets `cap_dac_override` and `uinput` is loaded, without which snippets and paste silently do not work
- the eight runtime dependencies stay with the package manager on Arch, where the names are known; on Fedora and Debian, where guessing them across releases is its own source of breakage, the binary is asked what it cannot resolve and the missing libraries are named

x86_64 only, since that is all upstream builds; other architectures are told to fetch the AppImage rather than failing halfway through.

## Uninstall

Losing the package manager means losing its file tracking, so `uninstall.sh` does that by hand — every path the tarball lays down under `/usr/local`, checked against the archive, all 58 files covered.

The generated theme, its regenerator and path unit go too. Nothing needs to be done about KRunner — the shortcut steal restores itself. The runtime dependencies deliberately are not removed: that list holds applications, not shared Qt and KDE libraries, and dropping `nodejs` or `qt6-svg` would take unrelated software with it.

## Tested

Live, end to end: service up, `Alt+Space` invoked over KGlobalAccel's D-Bus interface with the launcher coming up, the key taken from KRunner and handed back on exit, theme applied and regenerating when the scheme file changes.

The theme was checked against pixels rather than by eye — background `#131317` matches the theme exactly, the selected row `#2a292e` is the palette's `surfaceContainerHigh`, the accent reads `#c5c4ff` against `#c2c1ff`.

**Not tested:** the `/usr/local` install path itself. My test install went to `~/.local` because I had no way to authenticate for `sudo`; the tarball is relocatable so everything above still exercised, but `setcap`, `modprobe` and the `/usr/local` layout were verified by unpacking and inspection rather than by running the script. Worth a real run before merge.

### DIFF
--- a/scripts/06-services.sh
+++ b/scripts/06-services.sh
@@ -110,14 +110,12 @@ systemctl --user start ydotoold.service 2>/dev/null || \
     echo "  [INFO] ydotoold will start on next login."
 echo "  [OK]  ydotoold service configured."
 
-#  Vicinae (replaces KRunner)
-# KRunner is D-Bus activated, so stopping it is not enough — the next lookup
-# would start it straight back up. Masking is what actually keeps it down.
-echo "  Disabling KRunner..."
-systemctl --user mask plasma-krunner.service 2>/dev/null || true
-systemctl --user stop plasma-krunner.service 2>/dev/null || true
-pkill -x krunner 2>/dev/null || true
-
+#  Vicinae
+# KRunner is left alone. The shell binds Alt+Space through its own
+# GlobalShortcut, which takes the key off whoever holds it and hands it back
+# when the shell exits — KRunner included. Masking the service would disable it
+# outright and only be undone by the uninstaller, so a machine that simply
+# stopped running Caelestia would be left without it.
 if command -v vicinae >/dev/null 2>&1; then
     echo "  Enabling Vicinae launcher daemon..."
     # Both the AUR package and the upstream tarball ship a unit — /usr/lib for
--- a/scripts/06-services.sh
+++ b/scripts/06-services.sh
@@ -120,8 +120,11 @@ pkill -x krunner 2>/dev/null || true
 
 if command -v vicinae >/dev/null 2>&1; then
     echo "  Enabling Vicinae launcher daemon..."
-    # The package ships its own unit; only write one if it did not.
-    if [[ ! -f /usr/lib/systemd/user/vicinae.service && ! -f "$HOME/.config/systemd/user/vicinae.service" ]]; then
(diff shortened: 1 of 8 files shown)


0xSolanaceae:
we're going to keep krunner for now, thanks for the commit though!

PR #488, opened 2026-08-18
fix: keep keyboard layout indicator in sync with KDE

Klivan49:
The KbLayout service reads the active layout from org.kde.keyboard only
once at startup (and after switches initiated from the shell), while the
sidebar KbLayoutModel reads it only when the popout is created. Switching
layouts with Plasma's own shortcut therefore leaves the bar indicator and
the sidebar popout stale.

Changes:
- services/KbLayout.qml: poll the active layout every 750 ms while the
  shell is running (only when the process is idle)
- modules/bar/popouts/kblayout/KbLayoutModel.qml: follow the singleton
  via a Connections handler on activeIndexChanged

### DIFF
--- a/shell/modules/bar/popouts/kblayout/KbLayoutModel.qml
+++ b/shell/modules/bar/popouts/kblayout/KbLayoutModel.qml
@@ -4,6 +4,7 @@ import QtQuick
 import Quickshell.Io
 import Caelestia
 import Caelestia.Config
+import qs.services
 
 // TODO: handle this better later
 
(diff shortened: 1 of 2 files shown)


ladybug-me:
Hlo @Klivan49 
Great work

Polling Timer is to be avoided.
Please replace it with a Process running dbus-monitor.

Here is a working block:

```qml
    Process {
        id: dbusMonitor

        command: ["dbus-monitor", "--session", "type='signal',interface='org.kde.KeyboardLayouts'"]
        running: root.started
        
        stdout: SplitParser {
            onRead: text => {
                if (text.indexOf("layoutChanged") !== -1) {
                    if (!activeProc.running)
                        activeProc.running = true
                } else if (text.indexOf("layoutListChanged") !== -1) {
                    if (!layoutsProc.running)
                        layoutsProc.running = true
                }
            }
        }
    }
```

ladybug-me:
@Klivan49 
Please fix qml lint convention errors
Rest is ready to be merged

PR #507, opened 2026-08-21
fix: open launcher and window switcher on the active monitor

### LINKED ISSUE #442: Per monitor Notification and Bar visibility support
### What's the idea?

Most people use it for Discord/Browser, and it takes up some space. Also, there's a bug where pressing Meta to open search opens it on the second monitor no matter which one is in focus. I don't know how to reproduce it yet, so I haven't reported it.

The notifications duplicate on both monitors as well, which is pretty unpleasant.

### Why would it be useful?

Just quality of life

### LINKED ISSUE #513: Dock is opening on the wrong monitor
### Issue type

### What happened?

No matter which app on which monitor is in focus, dock is opening on the wrong one

#442 Would fix it

May be related to #507 

### Steps to reproduce

I honestly have no idea why it's happening or how to fix it

### Caelestia version

v2.3.1

### Distro

CachyOS

### Logs or screenshots

https://github.com/user-attachments/assets/758f31fa-2f4e-441d-b345-c21a16ed8d49

jedrikjames:

## What does this change?

Fixes the launcher and Alt+Tab window switcher always opening on the same monitor in multi-monitor setups.

The active drawer is now selected using KWin's current output when the shortcut is triggered, so it follows the cursor even when the cursor is over another application or the desktop. Launcher visibility is also kept exclusive across outputs, preventing a second launcher or window switcher from remaining open after moving between monitors.

Fixes #442, #513.

## Screenshots (if UI change)

#513: Only opened on the 2nd monitor

Fix:

https://github.com/user-attachments/assets/5bbd58c0-3811-49fc-a72a-0b560429a53f

## How did you test it?

- Rebuilt and installed the shell plugin on KDE Wayland with two monitors.
- Opened the launcher with the cursor over application windows, the desktop, and shell surfaces on both monitors.
- Moved the cursor to the other monitor while the launcher was open and confirmed the panel moved without leaving another instance open.
- Repeated the same check with Alt+Tab and confirmed the window switcher transferred between monitors.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

The monitor lookup uses KWin's `activeOutputName` instead of `QCursor::pos()`, since Wayland only updated the latter while the cursor was over a Quickshell surface. The existing focused-monitor lookup remains as a fallback if KWin does not return a matching output.

### DIFF
--- a/shell/modules/screenshot/regionSelector/RegionSelector.qml
+++ b/shell/modules/screenshot/regionSelector/RegionSelector.qml
@@ -3,6 +3,7 @@ import QtQuick
 import Quickshell
 import Quickshell.Hyprland
 import Quickshell.Io
+import Caelestia.Services
 import qs.components.misc
 import qs.services
 
@@ -26,7 +27,7 @@ Scope {
 
             required property var modelData
 
-            active: root.screenshotActive && modelData.name === Hypr.focusedMonitor.name
+            active: root.screenshotActive && modelData.name === KWinActiveWindowBridge.cursorOutputName()
 
             sourceComponent: RegionSelection {
                 screen: regionSelectorLoader.modelData
--- a/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
+++ b/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
@@ -45,6 +45,11 @@ QString KWinActiveWindowBridge::pendingFocusAddress() const {
     return m_pendingFocusAddress;
 }
 
+QString KWinActiveWindowBridge::cursorOutputName() const {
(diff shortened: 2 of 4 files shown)


ladybug-me:
Great work @jedrikjames 
Could you also apply this to RegionSelector (screenshot) ?
Make sure to pull latest commits from dev first

Its good to merge even now if you want

jedrikjames:
Thank you @ladybug-me 
Just committed 48f121a4b88c5ae43fe33f5120ba2f498d2e447c, should apply to RegionSelector now. 


https://github.com/user-attachments/assets/65d795f3-742c-481a-9847-ba9f763f1f77

ladybug-me:
Great work!
`KWinActiveWindowBridge::cursorOutputName` is way better then the previous fix.

PR #508, opened 2026-08-21
refactor: drop the screencast protocol and its KWin privilege, capture thumbnails inside KWin instead

WinTone01:
Removes `zkde_screencast_unstable_v1` and the KWin privilege it required, then puts real window previews back without it.

Two commits, meant to be read in order: the first takes the capability and the grant away, the second rebuilds the capability without the grant.

## Why

Live thumbnails were drawn by asking KWin for a PipeWire feed of each window. That protocol is privileged, so the installer wrote a `.desktop` override whose `X-KDE-Wayland-Interfaces` line existed for no reason other than to make KWin hand the shell a capability it would otherwise refuse. A standing grant like that outlives the feature that asked for it — and nothing ever removed it, so it kept applying to whatever `quickshell` binary later occupied the same path, including after uninstall.

It was also implicated in two problems: the crashes seen while hovering taskbar previews, and the screen-sharing conflict already documented in TROUBLESHOOTING 3.3.1, where a second client on the same protocol (Vesktop share, a camera through xdg-desktop-portal-kde) would freeze on some NVIDIA setups.

## What changed

**Removed** — `WindowScreencastRequest`/`OutputScreencastRequest` and the protocol XML; `ScreencastManager.qml`; the KPipeWire dependency it was the only user of, from every distro's package list; the `bar.livePreviews` setting and its Nexus toggle, which would otherwise be a switch wired to nothing; `assets/org.quickshell.desktop`, which requested nothing else.

**Kept** — `org_kde_plasma_window_management`. It is the other half of that same privilege line and it is what `KWinActiveWindowBridge` is built on (window list, focus, close, minimise), so the taskbar, window switcher and overview all stop working without it. Only the screencast half is gone.

**Replaced with** — capture inside the workspace-tracker KWin effect. That code is already part of the compositor, so it reaches the scene graph directly and needs no permission from it: `EffectWindow` → `windowItem()` → `scene()` → `renderer()`, rendered into an offscreen `GLFramebuffer` and read back as a PNG. Same path KWin's own screenshot effect takes. The only channel out is a D-Bus object on the service KWin already owns — no new bus name, no policy file.

`org.kde.KWin.ScreenShot2` is deliberately not used: it gates callers on `X-KDE-DBUS-Restricted-Interfaces`, i.e. exactly the kind of standing grant this PR removes.

Captures are queued and drained from `postPaintScreen` rather than answered inline on the D-Bus call, so they always run with a current GL context inside a settled frame; reading the scene from an arbitrary event-loop callback can hand back a black or half-drawn texture. Callers get a delayed reply. The queue is capped — past that, requests are answered empty, which is already the "show the icon" path.

Shell side is a declarative `WindowThumbnail` type backed by C++. That is deliberate: `ScreencastManager` kept a JS dictionary of refcounted streams and was asked for one from `Component.onCompleted`, which ran object creation inside `QQmlObjectCreator::finalize` and crashed the V4 engine (#496 was a mitigation for exactly this). Here a consumer states which window it wants and binds to `source`; setting a property only arms a timer and the call is async, so there is nothing to defer out of incubation and no refcount to leak. A short shared TTL lets two surfaces showing the same window reuse one capture, and an `active` property keeps offscreen overview pages from paying for readbacks nobody sees.

Minimised, hidden and deleted windows are refused rather than captured — they render as fully transparent, which the shell cannot tell apart from a real frame. All four surfaces keep an icon fallback for those cases.

`10-autostart.sh` deletes the stale override on upgrade, and `uninstall.sh` now removes both — neither was cleaned up before.

## Verified

Directly, with screenshots: the raw D-Bus capture (1270×907 window → 512×366 PNG, correct orientation and aspect, real content), the dock hover popup, and the overview. Both surfaces render legible window content, correctly proportioned and within bounds.

Not seen on screen: the Alt+Tab switcher and windowinfo. They use the identical `WindowThumbnail` + `Image` + icon-fallback pattern and the shell loads with no QML errors, but I could not trigger them to photograph.

## Things a reviewer should know

- **A new build of the effect does not take effect until the session restarts.** `unloadEffect`/`loadEffect` destroys and recreates the C++ object but does not `dlclose` the library, so KWin keeps running the old code. Confirmed here. The install flow already installs with sudo and picks it up at next login, but it will look like "the change did nothing" to anyone testing in place.
- **Still frames, not live video.** That trade came with dropping the protocol. `refresh()` exists for callers that want a newer one; periodic refresh was left out rather than paying for steady readbacks inside KWin.
- **The effect compiles against KWin's internal headers**, so it needs rebuilding on KWin updates. Already true of the existing workspace-tracker effect — not a new constraint, but this adds more surface to it.


WinTone01:
Closing this: the premise did not survive testing.

Removing the protocol worked, and capturing inside the workspace-tracker effect did produce real window content with no privilege at all. What it could not produce was a preview that moves. The cost is not the PNG encode — that was moved off the render thread and the compression turned off, and it was still slow. It is `glReadPixels`: a GPU→CPU→GPU round trip per frame, on the thread that drives compositing. Better transport removes the encode, not the readback. Avoiding the readback means exporting a dmabuf from the effect and importing it in the shell as an EGLImage, which is reimplementing `zkde_screencast_unstable_v1` by hand, against a QML stack with no dmabuf-importing item.

Two other routes were checked and neither is available:

- **`ext-image-copy-capture-v1`**, with `ext_foreign_toplevel_image_capture_source_manager_v1` as the source, would be exactly right and needs no privilege. KWin 6.7.4 does not implement it — there is no trace of the interface in the binary.
- **xdg-desktop-portal ScreenCast** is present (v5, window sources included) but puts an interactive picker and a consent dialog in front of every source. A shell cannot enumerate and thumbnail every open window through it.

More importantly, the premise itself was wrong. The privilege was suspected of causing the crashes, and that was never demonstrated — the reproducer came back 0/20 with the fix removed as well as with it in place, so it proved nothing either way. The one crash that *was* diagnosed had nothing to do with the protocol: `ScreencastManager.qml` mutated a JS dictionary of refcounted streams from `Component.onCompleted`, which runs inside `QQmlObjectCreator::finalize`, and V4 segfaulted in `insertMember`.

So the protocol stays and that bookkeeping goes. Replaced by #510: a declarative C++ `WindowStream` type, where there is no function left to call at the wrong time.

PR #511, opened 2026-08-22
fix: register the notification server before user applications start

WinTone01:
Some notifications arrived as the application's own popup, in a corner the shell does not control, and which applications did that varied between logins.

> Split at review request: the Panels change that was in here is now #544, which stands on its own. This PR is the autostart phase alone.

## What is actually happening

Applications decide **once, when they start**, whether a notification server exists. One that finds none falls back to drawing its own popups for the rest of the session, ignoring every setting on this side. That is why the effect is per-application and sticky rather than per-notification — it looks random, but it is decided at each application's launch.

The shell's autostart entry was `X-KDE-AutostartPhase=2`. That is the Applications phase — the same phase the user's own autostarted applications run in, with no ordering between them. Whether a given application saw a notification server was settled by whichever won the race that login.

Phase 1, DesktopServices, completes before any of them begin.

## Steps to reproduce

**The gap and what it costs — no logout needed:**

```bash
# stop the shell
pkill -x quickshell

# nobody owns the service now
qdbus6 org.freedesktop.DBus /org/freedesktop/DBus \
  org.freedesktop.DBus.GetNameOwner org.freedesktop.Notifications
#   Error: org.freedesktop.DBus.Error.NameHasNoOwner

# send one and time it
time notify-send test test
```

| | result |
|---|---|
| shell stopped | hangs **~85s**, then `Failed to show notification: Error calling StartServiceByName for org.freedesktop.Notifications: Timeout was reached` |
| shell running | returns in **~28ms** |

So a call landing in that window does not merely go unstyled — it blocks the caller for over a minute. The hang is `/usr/share/dbus-1/services/org.kde.plasma.Notifications.service` (`Exec=/usr/bin/plasma_waitforname org.freedesktop.Notifications`): D-Bus activates it, it waits for plasmashell to claim the name, plasmashell never does, and it times out — `WaitForName: Service was not registered within timeout` in the journal.

**Why the window exists at login:**

```bash
grep -H 'X-KDE-AutostartPhase' ~/.config/autostart/*.desktop
```

On this machine, before the change, `caelestiashell.desktop` was phase 2 and the other entries — Ferdium, EasyEffects, a download manager — had no phase line at all, which also means 2.

## Being straight about what is measured

The gap and the 85s failure above are measured. The "keeps drawing its own for the rest of the session" part comes from observing the affected applications and from restarting one fixing it — I did not instrument any application to confirm it caches the lookup at startup. Happy to reword the commit message to claim only the measured part if you would rather.

## Scope

Phase 1 closes the window at login, before applications start. It does nothing for a gap opened later — if the shell exits mid-session, the same failure is there until it re-registers. Worth knowing, but a different problem from the one this fixes.

**Applications already running have made their decision** and need restarting once; from the next login it takes care of itself.

### DIFF
--- a/scripts/10-autostart.sh
+++ b/scripts/10-autostart.sh
@@ -59,6 +59,13 @@ exec stdbuf -oL -eL "$QUICKSHELL_PATH" -d -n -p "\$HOME/.config/quickshell/caele
 EOF
 chmod +x "$HOME/.local/bin/caelestia-autostart.sh"
 
+# Phase 1 (DesktopServices), not 2 (Applications). The shell registers
+# org.freedesktop.Notifications, and applications decide once, when they start,
+# whether a notification server exists -- one that finds none draws its own
+# popups for the rest of the session, in its own corner, ignoring every setting
+# here. In phase 2 the shell starts alongside the user's autostarted apps with
+# no ordering between them, so which apps end up talking to it is a coin toss
+# per login. Phase 1 finishes before any of them begin.
 cat > "$AUTOSTART_DIR/caelestiashell.desktop" << EOF
 [Desktop Entry]
 Type=Application
@@ -69,7 +76,7 @@ Icon=quickshell
 Hidden=false
 NoDisplay=false
 X-GNOME-Autostart-enabled=true
-X-KDE-AutostartPhase=2
+X-KDE-AutostartPhase=1
 X-KDE-Wayland-Interfaces=zkde_screencast_unstable_v1
 EOF
 echo "  [OK]  Quickshell autostart created."


ladybug-me:
@WinTone01 Could you provide steps to reproduce the issue for fix1?

WinTone01:
Sure. The part that is directly measurable is the window where nothing owns `org.freedesktop.Notifications`, and what a notification call does inside it.

### 1. The gap, and what it costs (no logout needed)

```bash
# stop the shell
pkill -x quickshell

# nobody owns the service now
qdbus6 org.freedesktop.DBus /org/freedesktop/DBus \
  org.freedesktop.DBus.GetNameOwner org.freedesktop.Notifications
#   Error: org.freedesktop.DBus.Error.NameHasNoOwner

# send one and time it
time notify-send test test
```

Measured here:

| | result |
|---|---|
| shell stopped | hangs **~85s**, then `Failed to show notification: Error calling StartServiceByName for org.freedesktop.Notifications: Timeout was reached` |
| shell running | returns in **~28ms** |

The hang comes from `/usr/share/dbus-1/services/org.kde.plasma.Notifications.service`, which is `Exec=/usr/bin/plasma_waitforname org.freedesktop.Notifications`. D-Bus activates it, it waits for plasmashell to claim the name, plasmashell never does, and it gives up — visible in the journal as `plasma_waitforname[...]: WaitForName: Service was not registered within timeout`.

So during that window a notification is not merely unstyled. It blocks the caller for well over a minute and then fails.

### 2. Why that window exists at login

```bash
grep -H 'X-KDE-AutostartPhase' ~/.config/autostart/*.desktop
```

On this machine, before the change:

```
caelestiashell.desktop:X-KDE-AutostartPhase=2
```

and the other entries — Ferdium, EasyEffects, a download manager — have no phase line at all, which also means 2. Phase 2 is the Applications phase, so the shell and the user's applications are started together with no ordering between them. Which of them queries the service before the shell has registered is decided per login.

To watch the race rather than infer it: log in, and check the shell's log for

```
quickshell.service.notifications: Could not register notification server at
org.freedesktop.Notifications, presumably because one is already registered.
```

and compare the ownership timestamps in `journalctl --user -b | grep -i Notifications`.

### 3. What the user actually reports

Applications that hit the window draw their own popups instead, in their own corner, and keep doing it for the rest of the session — the shell's own popups land where the config says (bottom-right, for a bottom bar; confirmed with `notify-send` while the shell holds the name). Restarting just the affected application, with the shell already up, sends its notifications through the shell again.

**Being straight about what I did and did not verify:** the gap and the 85s failure above are measured. The "keeps drawing its own for the rest of the session" part comes from observing the affected applications and from restarting one fixing it — I did not instrument any application to confirm it caches the lookup at startup. If you would rather the PR only claim the part that is measured, I am happy to reword the commit message.

### Scope

Phase 1 closes the window at login, before applications start. It does nothing for a gap opened later — if the shell exits mid-session, the same 85s failure is there until it re-registers. Worth knowing, but that is a different problem from the one this fixes.

The second change in this PR is unrelated to this one; it is the duplicated corner calculation in the auto-avoidance drifting from `notifAutoPosition`. Happy to split it into its own PR if you would prefer them separate.

ladybug-me:
@WinTone01 Appreciate the effort but starting quickshell during `Phase 1` does not seem to be a good option in my opinion.

I will put this in review but it would be better if some other alternative is found.

You may create a different PR for the `Panel.qml` fix.

0xSolanaceae (changes requested):
This is a super important fix, so I appreciate you throwing this in @WinTone01. If you could just split this into two PR's. The one with the Panels fix I can merge instantly

WinTone01:
Split done — @0xSolanaceae the Panels half is #544, on its own and nothing else in it.

This PR is now the autostart phase change alone: one line in `scripts/10-autostart.sh`, phase 2 to phase 1.

@ladybug-me the reproduction steps are in the description now rather than buried in a comment, along with what is measured (the ~85s hang while nothing owns the name) versus what is inferred (that applications cache the lookup at startup). I have not instrumented an application to prove the second part, so if you would rather the commit message only claimed the measured half, say so and I will reword it.

0xSolanaceae (approved):
Appreciate it!

PR #540, opened 2026-08-27
feat: per-monitor workspaces in the bar and overview, and cross-screen window dragging

WinTone01:
KWin gives each output its own current desktop when `PerOutputVirtualDesktops` is on, but the shell read the one D-Bus exposes — a single global value that follows whichever screen is focused. Every bar and every overview therefore showed the focused screen's desktop, and appeared to switch on all of them when the pointer crossed to another monitor.

## Per screen

Scoped to the screen it belongs to: the desktop a bar and an overview open on, the windows in the grid, the window icons in the bottom strip, and which desktops that strip counts as occupied. The window data already carried an `output` field, so most of this is a filter rather than new plumbing.

The desktop a screen is on can only be read per output from inside the compositor, so the workspace-tracker effect reports it and `KWinWorkspaceState` exposes `activeByOutput`. Switching goes the same way — D-Bus can only set the active output's desktop, which is wrong once two screens each have one and both can ask. The effect gained `SetDesktop(output, desktop)` for that.

**Version skew is handled.** A rebuilt effect does not load until the session restarts, so a shell update routinely runs against an older one. Switching probes for `SetDesktop` once and falls back to the global setter if it is missing, and the socket format carries a magic so the old 12-byte record is still understood. Without either, a shell update would silently stop workspace switching until the next login.

## The overview opens everywhere

It is a place things get dragged across: a window can be moved to another monitor, or to a desktop that only exists there, and neither works if the destination still shows the desktop underneath. `Visibilities.setOverview()` drives every screen, and all the open and close paths go through it.

That exposed a conflict worth naming: each screen's overview switches desktops as its page changes, and through the global setter one screen's page sync moved whichever monitor the pointer happened to be over. Measured — with the screens on desktops 2 and 1, opening the overview dragged the second one to 2 as well.

## Dragging

- **To another monitor.** The edge areas paged workspaces the instant a drag touched them, which on a side with a neighbouring screen swallowed exactly the gesture meaning "put this over there". They now wait for the drag to rest: pass through and you reach the next monitor, stop and you page, and holding keeps paging. Moving a window between screens has no unprivileged path — plasma-window-management does desktops, not outputs — so the effect gained `SendToOutput` too.
- **The drag does not vanish at the edge.** The card belongs to the overview it started in and is clipped there, so the second half of the gesture used to happen against nothing. The position is published and the screen the pointer reaches draws the same live preview, at the size the card had.
- **Onto a workspace thumbnail** a card collapses into the application's icon and expands back when pulled away, GNOME-style. Not over the workspace it is already on, where the shrink would be announcing a move to where it already is.
- **Out of the strip** an icon opens into the window it stands for. Driven by how far it has been lifted, on two thresholds — following the slots the drag passed over made it flip at every boundary, and each flip tore down and rebuilt the screencast.

One stream claim underpins the last two: a node feeds one consumer, so a preview outside the grid says it wants the stream and the card that holds it lets go, otherwise the new one just draws black.

## Verified

By driving a real pointer through `/dev/uinput` rather than by inspection — each of these was watched on screen or measured:

- desktops diverge per output (`{DP-1:2, HDMI-A-2:1}`) where the global value reports one
- a window dropped on the second monitor arrives there (`out=DP-1` → `out=HDMI-A-2`), and passing through the edge no longer changes the workspace
- holding at an edge advances one workspace, then another, and stops when released
- an icon dragged from the first thumbnail to the third moves that window to desktop 3
- a card reads as a full preview over its own workspace and as a bare icon, with no buttons on it, over another
- an icon held between the two thresholds is pixel-identical a second and a half apart

## Note on the base

Written before #510 and rebased onto it. The last commit ports the two preview sites this branch added onto `WindowPreview` rather than leaving the older inline plumbing behind — worth a look, since that is the part a rebase could have carried through quietly.

### DIFF
--- a/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
+++ b/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
@@ -3,7 +3,12 @@
 #include <KPluginFactory>
 #include <QStandardPaths>
 #include <QDir>
+#include <QDBusConnection>
+#include <QDBusError>
+#include <effect/effectwindow.h>
 #include <QDebug>
+#include <core/output.h>
+#include <cstring>
 #include <QStandardPaths>
 #include <QDir>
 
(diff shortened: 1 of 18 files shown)


0xSolanaceae (changes requested):
Resolve the merge conflicts and split the two features into two PR's

WinTone01:
Split into #548 (per-monitor workspaces) and #549 (cross-screen dragging), both rebased on current dev so the conflicts are resolved. #549 is stacked on #548. Closing this one.

PR #541, opened 2026-08-27
feat: add an EasyEffects quick toggle

### LINKED ISSUE #504: EasyEffects integration

WinTone01:
Closes #504.

EasyEffects exposes nothing worth talking to over D-Bus for this, so the process **is** the state: running means its chain is applied to the audio graph, not running means it is not. Both the native package and the Flatpak are probed, since either can be the one installed.

Two deliberate differences from the [end-4dots-kde](https://github.com/ladybug-me/end-4dots-kde) implementation the issue points at:

- **Availability hides the toggle rather than disabling it.** A button that cannot do anything is worse than no button, and the quick toggles already drop the VPN entry this way when no provider is enabled.
- **The state is read back after toggling, not assumed.** Starting and stopping are fire-and-forget; the reference sets `active` optimistically, so a launch that fails or a stop that does not take leaves the toggle showing something untrue until the shell restarts. Here a short confirmation re-reads the real process.

Registered in the quick toggles settings page as well, so it can be switched off like any other.

## Verified

On a machine with the native package installed and running:

| step | process | service |
|---|---|---|
| initial | running | `available=true active=true` |
| toggled off | gone | `active=false` |
| toggled on | running | `active=true` |

Not covered: the Flatpak path. The commands are there and mirror the reference, but this machine has the native package, so I have not seen them run.

### DIFF
--- a/shell/modules/nexus/pages/utilities/QuickTogglesPage.qml
+++ b/shell/modules/nexus/pages/utilities/QuickTogglesPage.qml
@@ -25,6 +25,7 @@ PageBase {
         { id: "dnd", label: qsTr("Do Not Disturb") },
         { id: "pauseWallpaper", label: qsTr("Pause Wallpaper") },
         { id: "nightlight", label: qsTr("Night Light") },
+        { id: "easyeffects", label: qsTr("EasyEffects") },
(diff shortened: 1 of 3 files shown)


0xSolanaceae (approved):
LGTM!

PR #544, opened 2026-08-27
fix: derive the notification avoidance corner from one place

WinTone01:
Split out of #511 at review request — this is the self-contained half.

The corner the notification auto-avoidance watches was worked out a second time inside the `popupCount` handler, and the two copies had drifted apart:

```qml
// notifAutoPosition
const h = barPos === "right" ? "left" : "right";
// the flip trigger
const defH = Config.bar.position === "left" ? "left" : "right";
```

For a left- or right-positioned bar these disagree. The avoidance then watched the corner *opposite* the one notifications actually appear in: it flipped away when the cursor was nowhere near them, and stayed put when the cursor was sitting on top of them — exactly backwards.

Now derived from `notifAutoPosition`, so there is one definition instead of two.

Top and bottom bars were unaffected, which is why this went unnoticed.


0xSolanaceae:
LGTM!

PR #548, opened 2026-08-28
feat: per-monitor workspaces

WinTone01:
First half of #540, split as requested. Rebased on current dev, so the conflicts are gone.

KWin already gives every output its own current virtual desktop, but the shell had no way to see it. The D-Bus `VirtualDesktopManager.current` exposes a single "current" that follows whichever output is focused, and setting it applies to that output too — so the bar, the overview and the workspace indicator all showed the focused screen's desktop on every monitor, and a switch meant for one monitor landed on whichever the pointer happened to be over.

The workspace-tracker effect can answer per output from inside the compositor, so it now reports the desktop for each one and takes a `SetDesktop(output, desktop)`. `KWinWorkspaceState` exposes that as `activeByOutput`, and `switchTo()` takes an optional output name, falling back to the D-Bus path when the effect is not installed.

Everything downstream is scoped to a screen: the overview knows which one it belongs to and filters windows by output, the bar's workspace strip reads its own screen's desktop, and the overview opens on every screen at once rather than only the focused one.

The other half — dragging windows between screens — is stacked on this one and goes up separately.

### DIFF
--- a/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
+++ b/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
@@ -3,7 +3,12 @@
 #include <KPluginFactory>
 #include <QStandardPaths>
 #include <QDir>
+#include <QDBusConnection>
+#include <QDBusError>
+#include <effect/effectwindow.h>
 #include <QDebug>
+#include <core/output.h>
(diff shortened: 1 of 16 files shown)


0xSolanaceae (approved):
Looks good, thanks!

PR #549, opened 2026-08-28
feat: drag windows between screens in the overview

WinTone01:
Second half of #540, split as requested. Stacked on #548 — that one should go first; this branch contains both commits, so the diff here is the whole thing until #548 lands.

The overview could already move a window to another workspace by dragging its card onto a workspace thumbnail, but not to another monitor. Each screen's overview is its own window and can only draw on its own screen, so a card dragged towards the next monitor was clipped at the edge and vanished — the drag was still running, but there was nothing to see and it read as dropping the window into nowhere.

The drag position is now published through `Visibilities`, so the screen the pointer has reached draws what is arriving, showing the same live preview the card was showing. A release outside the origin screen moves the window there via a new `sendToOutput()`: plasma-window-management can move a window between desktops but not between outputs and there is no D-Bus surface for it, so it goes through the workspace-tracker effect.

Two smaller things fell out of it:

- Holding a drag against a screen edge now waits before paging to the next workspace, so travelling towards the next monitor no longer pages one on the way past. Pass through and you reach the monitor, stop and you page.
- A card held over a workspace thumbnail collapses into the application's icon, so the target reads as a target rather than the card floating over it.

### DIFF
--- a/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
+++ b/shell/kwin-effects/workspace-tracker/workspace_tracker.cpp
@@ -61,6 +61,27 @@ KWin::LogicalOutput* WorkspaceTrackerEffect::findOutput(const QString& name)
     return nullptr;
 }
 
+void WorkspaceTrackerEffect::SendToOutput(const QString& uuid, const QString& output)
+{
+    KWin::LogicalOutput* target = findOutput(output);
+    if (!target) {
+        return;
+    }
+
+    const QUuid wanted(uuid);
+    if (wanted.isNull()) {
+        return;
+    }
+
+    const auto windows = KWin::effects->stackingOrder();
+    for (KWin::EffectWindow* window : windows) {
+        if (window && window->internalId() == wanted) {
+            KWin::effects->windowToScreen(window, target);
+            return;
+        }
+    }
+}
+
 void WorkspaceTrackerEffect::SetDesktop(const QString& output, int desktop)
 {
     if (desktop < 1) {
--- a/shell/kwin-effects/workspace-tracker/workspace_tracker.hpp
+++ b/shell/kwin-effects/workspace-tracker/workspace_tracker.hpp
@@ -3,6 +3,7 @@
 #include <effect/effect.h>
 #include <effect/effecthandler.h>
 #include <QLocalSocket>
+#include <QUuid>
 #include <QPointer>
 #include <QObject>
 #include <QPointF>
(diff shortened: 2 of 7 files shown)


ladybug-me (changes requested):
@WinTone01 The dragging between pages does not seem to work. Your intended timer does fire off and changes the page successfully but the dragging item moves back to its previous workspace.

Test this with single monitor

WinTone01:
Thanks, that was a real bug — `syncPage()` was fighting the drag.

The page a drag has reached is chosen by the drag, but `activeWsId` only catches up once the compositor has switched and the workspace-tracker payload has come back over its socket. Any `activeWsId` notification arriving in that gap still carries the desktop the drag started on, and `syncPage()` acted on it and snapped the view back to that page. The drop then compared the page it had been pulled back to against the one the window was already on, found them equal, and did nothing — so the card flew home and the window stayed put. The timer paging correctly and the drop being ignored is exactly that.

`syncPage()` now stands aside while a drag is running and runs once when it ends, so the view still settles on the real desktop afterwards.

I could not reproduce it here — my setup is multi-monitor, where the tracker payload evidently lands before the next notification does. So this is reasoned from the code rather than confirmed against the failure, and I would rather you re-test it on the single-monitor case than take my word for it. If it still reverts, the next thing I would check is whether `listView.currentIndex` is what you expect at release time.

WinTone01:
Rebased on current dev now that #548 is merged, so this is just the dragging change: 2 commits, 7 files.

You were right, and it was worse than one bug. Three things were in the way, all of which only appear because the gesture requires holding the pointer still:

- Qt re-resolves which DropArea a drag is over on drag movement, not when the scene moves underneath it — which is exactly what paging does. So the drag stayed registered against the page it started on, and that page's `DropArea` accepted the drop whether or not it had anything to do: `drop.accept()` sat outside the check for the window already being on this page. The card's release path saw a handled drop and never ran. That is the "page changes, window stays" you hit. A page now accepts only when it actually takes the window in.
- Switching the compositor's desktop cancels the pointer grab the drag rides on, and that switch follows a page change by 50ms. The drag died on the first page turn, so holding on could never carry the window further — it was dropped where it had just arrived. The desktop is now left alone for the length of the drag and catches up when it ends. `syncPage()` has to stand aside meanwhile, or it pulls the view back a page at a time.
- The stream gate keeps previews to the current page and its neighbours. A card carried two pages from home fell outside it and collapsed into its icon mid-gesture. That rule is about pages you cannot see, so it no longer applies to a card in your hand.

I also widened the edge band from 60px to 140px. On an edge shared with another monitor the pointer glides across instead of stopping, so 60px left almost nothing to aim at; the dwell, not the width, is what separates passing through from paging.

This is confirmed working now, not reasoned: reproduced the original failure on hardware, instrumented the drop path and the drag lifetime to find each cause, and re-tested after each fix. Holding at the edge now pages one workspace at a time for as long as you hold, and the window arrives where you let go.

Unrelated, spotted while in here: `Behavior on contentX` in `WindowGrid.qml` references `rawSwipeOffset` from outside its scope, so it throws `ReferenceError` on every page change and the animation duration never applies. It is on dev too. Happy to send it separately rather than fold it in here.

PR #550, opened 2026-08-29
feat: add UI translations with English and Turkish

### LINKED ISSUE #514: Add internationalization (i18n) - Spanish translation support
### What's the idea?

Add internationalization (i18n) support to Caelestia Shell so the UI elements, Nexus panel, and widgets can be translated into other languages (e.g., Spanish).

Currently, UI text strings are hardcoded in English within the `.qml` files. Non-English users on distros like Nobara/Debian/Fedora have their OS in their native language, but the Caelestia shell elements remain exclusively in English.

### Why would it be useful?

Implement a translation mechanism (such as Qt's `qsTr()` / `.ts` translation files) so the community can contribute localization files for different languages.

WinTone01:
Closes #514.

## What does this change?

The shell already had ~1600 `qsTr()` calls, but nothing to resolve them against, so the UI stayed English no matter the system locale. This adds the missing pipeline and the first two languages.

- `Translations` singleton (`shell/plugin/src/Caelestia/translations.{hpp,cpp}`) finds the matching catalogue, installs a `QTranslator` and calls `QQmlEngine::retranslate()`, so switching language applies **without restarting the shell**. Lookup order: `$CAELESTIA_TRANSLATIONS_DIR`, then `translations/` next to the running `shell.qml` (checkout or install tree), then the path baked in at build time.
- New `general.language` option (`"system"` by default) plus a real language picker in Nexus -> Language & region, replacing the read-only "the shell follows your system locale" row. Languages with an installed catalogue show up on their own.
- `shell/translations/caelestia_tr.ts`: complete Turkish translation, no unfinished entries. `caelestia_en.ts` is the source-language catalogue and doubles as the template contributors copy.
- CMake compiles the catalogues with `lrelease` and installs them next to the shell. Without Qt's Linguist tools the build just warns and ships English.
- `scripts/update-translations.sh` re-extracts strings (`scripts/update-translations.sh` for all, `scripts/update-translations.sh es` to start a new language), and `.github/docs/translations.md` documents the whole thing.
- Wrapped the ~64 user-facing literals that were still missing `qsTr()` (KMYCSettings, PolkitDialog, WeatherTab, AiAssistant, ...). Technical placeholders like `ghp_...` were deliberately left alone.

Spanish is not included - the point is the mechanism, so the community can add a language with one command.

## How did you test it?

- Built the plugin clean, then drove the `Translations` singleton standalone with `qml`: catalogue discovery, loading `tr` (`About` -> `Hakkinda`, `Chroma Multiplier` -> `Kroma Carpani`), and falling back to the source strings on `en`.
- `check_qml_conventions.py` (no new violations, 5 resolved), `check_qml_imports.py`, `check_config_references.py`, `test_repo_integrity.py` all pass.
- Turkish is my native language, so the translation is reviewed rather than machine-dumped.

## Type

- [x] New feature

## Notes for reviewers

- English needs no catalogue: it is the source language, so `caelestia_en.ts` intentionally ships with empty translations and selecting English simply removes the translator.
- `qsTr()` context is the QML file's base name, which is why the same word can be translated differently per file.
- This branch and `perf/updater-rebuilds` were kept textually apart on purpose - they merge cleanly in either order (verified both directions). If the perf PR lands first, the `lrelease` install in `08-build-shell.sh` can switch from `sudo` to that PR's `caelestia_sudo` helper.


WinTone01:
On the difference from #589 — they solve the same problem in two different places, so it is worth separating the mechanism from the size.

**Mechanism.** #589 adds an `I18n.tr()` singleton that looks strings up in a JSON map keyed by the English source text, and rewrites `qsTr(` to `I18n.tr(` across the tree. This PR keeps `qsTr()` and gives it what it was missing: a `QTranslator` installed from compiled `.qm` catalogues, with `QQmlEngine::retranslate()` for live switching. #589's premise — that Quickshell has no `QTranslator` integration — is right; this PR is an implementation of that integration rather than a way around it.

**Size**, since that is the part that looks lopsided. Of this PR's diff, 9796 lines are `caelestia_tr.ts`, the Turkish translation itself. The hand-written part is ~700 lines, and only ~120 of those modify existing files, across 20 of them. #589 is +2981/-1564 but ~170 of its 173 files are existing source files being rewritten. So by added lines #589 is smaller; by how much existing code has to be reviewed and re-checked, this one is. I have just dropped `caelestia_en.ts` from this branch — 9794 lines of empty translations kept only as a copy-from template, which `update-translations.sh` generates on demand — taking the PR from +20293 to +10499.

**Where #589 is genuinely better:** it needs no C++ and no Qt Linguist tools at build time. This PR needs `lrelease` to compile catalogues, and warns and ships English if it is absent.

**What `qsTr()` gets you that a source-string key does not:** plurals, per-file context so the same English word can differ by place, and — the one that matters for upkeep — `lupdate` marks an entry unfinished when its English source changes, whereas a source-string key silently misses and falls back to English with no signal that the translation went stale.

**On multilang:** here a language is `scripts/update-translations.sh es`, then Qt Linguist; `.ts` is also the format Weblate and Crowdin read natively, so community translators need no repo access. In #589 it is a JSON file plus a line in `I18n.languages`, hand-edited.

I am not arguing #589 should be closed — if the build-time dependency on Qt tooling is the thing you want to avoid, that is a real reason to prefer its approach, and I would rather you pick on that than on the line count.

Two things still on me here: this branch conflicts with current dev in `shell/lockscreen.qml` and `shell/plugin/src/Caelestia/CMakeLists.txt`, and the catalogue needs `lupdate` re-run against dev to pick up strings added since. Say the word and I will do both.

shell/modules/nexus/pages/wallandstyle/KMYCSettings.qml
```diff
@@ -391,23 +391,23 @@ PageBase {
...
-                text: "Klassy Windeco Outline"
-                subtext: "Tint Klassy Window Decoration window outline (Reloads KWin)"
+                text: qsTr("Klassy Windeco Outline")
+                subtext: qsTr("Tint Klassy Window Decoration window outline (Reloads KWin)")
                 checked: root.klassyWindecoOutline
                 onToggled: root.setOption("klassy_windeco_outline", checked ? "True" : "False")
             }
             ToggleRow {
                 Layout.topMargin: Tokens.spacing.extraSmall / 2
                 text: "KDE Rounded Corners Outline"
```
0xSolanaceae: Missed this one with qsTr

shell/plugin/src/Caelestia/translations.cpp
```diff
@@ -0,0 +1,241 @@
...
+void Translations::reload() {
+    refreshAvailable();
+
+    const QStringList dirs = searchPaths();
+    auto* translator = new QTranslator(this);
+
+    QString loaded;
+    for (const QString& code : candidates()) {
+        if (isSourceLanguage(code)) {
+            break;
```
0xSolanaceae, line 160: This `break` exits the whole candidate loop as soon as a source-language code is hit, so `reload()` will never attempt - let alone load - an `en_*` catalogue. But `refreshAvailable()` below seeds `"en"` and then adds every `caelestia_*.qm` it finds on disk without the same source-language filter, so `available()` can advertise a catalogue (e.g. `en_GB`) that `reload()` structurally can never activate: the picker would list "English (GB)", but selecting it would always fall back to `"en"` source strings.

Latent today since only `tr` ships, but the two functions already disagree by design. Line 30's comment says English is the source language and never needs a catalogue - if that's the rule, `refreshAvailable()` should filter `en*` codes out too, so the picker can't offer a language that can never load. (Switching this to `continue` alone wouldn't fix it - `en_GB` is still source, so it would be skipped either way.)

0xSolanaceae (changes requested):
two simple changes, thanks!

shell/modules/nexus/pages/wallandstyle/KMYCSettings.qml
WinTone01: Wrapped it, and the section header above it ("Window Decorations (Requires Plugins)") was bare too. Both translated.

shell/plugin/src/Caelestia/translations.cpp
WinTone01: Fixed the way you suggested: refreshAvailable() now skips source-language codes when scanning, so the picker offers the seeded "en" entry and nothing that reload() cannot activate.

Confirmed rather than reasoned. I drove the singleton standalone against a directory holding caelestia_tr.qm plus a caelestia_en_GB.qm. With the filter removed the available list came back carrying en_GB; with it in place the list is en and tr. So the disagreement you spotted was reachable, not only latent by design.

0xSolanaceae (approved):

PR #551, opened 2026-08-29
perf: stop the updater asking for passwords and hogging every core

WinTone01:
## What does this change?

A routine `caelestia-update` asked for the password several times and rebuilt more than it had to. On a desktop that shows up as repeated prompts and a session that stutters for the length of the build.

**Privileges**

- New `scripts/lib/privileges.sh`: one shared escalation helper that escalates **lazily**, reuses a single credential for the whole run, and works with no controlling terminal (askpass, then pkexec).
- `update.sh` no longer primes sudo up front. An update with no root work to do now asks for nothing at all; the first step that genuinely needs root prompts once.
- The package manager only runs for packages that are actually missing, instead of an unconditional `pacman -S --needed ...` on every update.
- The workspace-tracker root install is skipped when the freshly built effect is already the installed one, and KWin is only reconfigured when it changed. (`cmp` cannot be used here - the install step strips the binary - so the last install is tracked with a stamp.)
- `06-services.sh` skips its root block once the udev rule, `input` group membership and `/dev/uinput` permissions are in place.
- `ksshaskpass` added to the dependency lists so a GUI-launched update can prompt once instead of polkit prompting per privileged command.

**Build cost**

- The toolchain fingerprint now records only the Qt/CMake *feature* version. A patch release (6.11.2 -> 6.11.3) was wiping the build directory and forcing a full rebuild for nothing. Existing stamps are normalised, so this change does not itself trigger one.
- Builds run with `nproc-1` jobs at `nice 10` with a load limit, overridable with `CAELESTIA_BUILD_JOBS`.

## How did you test it?

Measured on a live install before changing anything: a no-op rebuild is already 0.05 s (`ninja: no work to do`), so the "it recompiles everything again" reports come from the build directory being wiped - a full rebuild after a Qt patch bump was 135 files, 71 s, 485% CPU.

- Verified the normalised stamp matches an existing pre-change build directory, so no forced rebuild on upgrade.
- Ran the workspace-tracker up-to-date check against the real installed effect: no stamp -> installs, after recording -> skips.
- Ran the missing-package check against a real system: correctly reports only what is absent.
- Exercised the guarded `06-services.sh` path end to end; it skipped the root step as intended.
- `bash -n` on every touched script, plus `check_shell_quality.sh`, `check_embedded_bash.py` and `test_repo_integrity.py`.
- Confirmed the tuned build command (`-j`, `nice`, `-l`) works with both the Ninja and Makefiles generators.

## Type

- [x] Bug fix
- [x] Other (performance / UX)

## Notes for reviewers

- The strict-mode CI rule now skips `scripts/lib/`: sourced helpers must not impose `set -e` on their callers, and `update.sh` deliberately runs without it.
- Escalation being lazy is the behavioural change worth a second look. `09-system-tweaks.sh` keeps its own never-prompt helper, so its `chsh` step still degrades to a warning rather than a prompt.
- This branch and `feat/i18n-translations` were kept textually apart on purpose - they merge cleanly in either order (verified both directions).


0xSolanaceae:
@WinTone01 super appreciate you throwing this in, if you could fix the conflict and update it to the new installer I will merge this now

WinTone01:
Thanks! Rebased onto the new installer and the conflict is gone - green and mergeable now.

What changed for the installer redesign:

- `lib/privileges.sh` is now sourced right after `lib/log.sh` in every script that uses it, and it reuses the shared `err` marker so the TUI's WARN/ERR scanning still works.
- The `-n` probe and the `-S` password feed go through `/usr/bin/sudo` directly. Under the TUI, `sudo` on PATH is the wrapper that forces `-A`, which breaks both modes - the same reason 09-system-tweaks.sh reaches for the real binary in bd4bafd. Verified with a fake wrapper on PATH: the probe and the `SUDO_PASS` path both bypass it, and the `SUDO_PASS` case (the one the TUI sets up) still works.
- Update runs `update.sh` on the real terminal before the TUI's sudo phase, so nothing there sees the wrapper or `SUDO_PASS`. That path keeps the lazy behaviour: an update with no root work asks for nothing at all.

One thing worth flagging separately: `scripts/lib/log.sh` currently fails the repo's own strict-mode check on dev.

```
$ bash .github/scripts/check_shell_quality.sh
[ERR]   scripts/lib/log.sh is missing 'set -euo pipefail' (required for scripts in scripts/)
```

`git ls-files 'scripts/*.sh'` matches nested paths, so the rule reaches into `scripts/lib/`. This PR already exempts that directory - sourced helpers must not impose `set -e` on their callers, and `update.sh` deliberately runs without it - so merging this fixes that too.

0xSolanaceae (changes requested):
`caelestia_sudo_quiet` (privileges.sh:126) is documented and defined, but nothing in this PR calls it. It's solving a problem the spec doesn't have. Either drop it or wire it into something.

caelestia_real_sudo also re-implements the `/usr/bin/sudo` bypass that 09-system-tweaks.sh's run_sudo_non_interactive already handles, the comment literally says "same reasoning as 09-system-tweaks." Now there are two bypass implementations to keep in sync. 09 should've been migrated here too, not left for later.

WinTone01:
Both fair - fixed in 81c6491.

`caelestia_sudo_quiet` and the duplicated bypass turned out to be the same problem, so one change closes both: the SUDO_PASS branch moved into `caelestia_sudo_quiet`, and `09-system-tweaks.sh` now sources the helper and calls it. `run_sudo_non_interactive` is gone, there is one `/usr/bin/sudo` bypass left in the tree, and the helper has a real caller instead of being an unused API. Dropped the stale "same reasoning as 09-system-tweaks" comment too, since 09 is no longer a separate implementation to reason about.

Behaviour is unchanged for 09: cached credentials or the exported password, never a prompt. Re-checked both modes with a fake `-A` wrapper on PATH - each one still reaches the real binary - and `grep -rn run_sudo_non_interactive` comes back empty.

0xSolanaceae (approved):
Looks good, thanks!

PR #552, opened 2026-08-29
fix: make the markdownlint config actually apply

WinTone01:
## What does this change?

`.markdownlint-cli2.jsonc` held its rule settings at the top level - that is the `.markdownlint.jsonc` schema. markdownlint-cli2 reads its own config file as an **options object**, so `"line-length": false` and every other setting were silently ignored and the default rule set ran instead.

That is why PRs touching markdown pick up a wall of `MD013/line-length` annotations on prose the repo deliberately does not hard-wrap, even though the config has been disabling that rule all along.

- Move the settings under `"config"`, where markdownlint-cli2 looks for them
- Pin `table-column-style` to `compact`, which is how tables are already written here
- Clear the violations that were left over in the two files the CI globs actually cover (`.github/README.md`, `.github/CONTRIBUTING.md`): list and fence spacing, one table delimiter row, a stray double blank line, and a heading that skipped from h1 to h3

## How did you test it?

Reproduced first with a throwaway file: a 97-character line was flagged with the config as it was, and clean once the settings moved under `"config"`.

Then ran the exact CI globs locally:

- before: MD013 on every unwrapped line, plus 18 other issues
- after: `Summary: 0 issues in 0 files`

## Type

- [x] Bug fix
- [x] Docs

## Notes for reviewers

- The README screenshot URL stays bare with a targeted `markdownlint-disable-next-line MD034`: a bare URL is what GitHub turns into an inline video player, so linking it would break the embed.
- The header subtitle became `<h3>` instead of `###` to satisfy MD001 without changing how it renders (`no-inline-html` is already off in the config).
- `.github/docs/**` is still outside the workflow's globs, so those files are not linted. With the config working they come to ~240 mechanical issues (table pipe spacing, `*` vs `-` bullets); worth a separate pass if you want that directory covered too.
- Merges cleanly with #550 and #551 in any order - verified all three together.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -22,7 +22,7 @@ Make your changes in the cloned repo, test them (see below), then open a PR. Tha
 ## Where stuff lives
 
 | Area | Directory | Tech |
-|------|-----------|------|
+| ------ | ----------- | ------ |
 | Shell UI (launcher, bar, notifications, etc.) | `shell/` | QML + Quickshell |
 | KWin plugin (window management, shortcuts) | `shell/plugin/` | C++ |
 | TUI installer | `installer/src/` | C++ |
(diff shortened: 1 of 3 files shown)


0xSolanaceae (approved):
LGTM!

PR #585, opened 2026-09-02
fix: stop the shell's audio capture from getting in other apps' way

### LINKED ISSUE #402: Error that causes Vesktop (discord) crash when using camera and/or screen share with audio
### Issue type

### What happened?

Configuration:
GPU: Nvidia Geforce RTX 4060
CPU: Intel Core i7-4790 (3.60hz)
OS: CachyOS
Kernel: 7.1.5-1-cachOS
RAM: 32GB 

I've been running into an issue when I use vesktop,

if I stream with audio or turn on my camera, it'll freeze vesktop and I'm forced to restart the app.
If I stream without audio, it'll work fine sometimes
Issue seems to be only with Vesktop (haven't tried regular discord) since I tried it on the browser and worked flawlessly.

it started once I downloaded caelestia

### Steps to reproduce

1. Download vesktop
2. join a call 
3. turn on camera or stream with audio

### Caelestia version

v2.2.2

### Distro

CachyOS

### Logs or screenshots

_No response_

WinTone01:
Towards #402.

## What does this change?

The screencast half of #402 already got a switch and a stream cap. The audio half had neither, and the most recent report in the thread is an audio one: @Lycham's OBS cannot capture, and starting or closing it makes the shell announce that the microphone and speaker changed when nothing did.

Verified on a running session with `pw-dump`: the shell holds a PipeWire capture of the sink monitor (`stream.capture.sink=true`, `node.passive=true`, state `running`) for as long as anything draws a visualiser - and `background.visualiser.enabled` defaults to true, so on a normal session that is always. Spectacle, which captures no audio, is unaffected; that matches what the thread reports.

- **`services.audioCapture` (default on), with a toggle in Services settings.** Off means the shell holds no capture stream at all - the audio-graph equivalent of `bar.livePreviews`. Users hitting this now have a supported way out instead of turning off the whole visualiser and hoping.
- **`Service` gains an `enabled` gate**, so the switch releases the stream without every consumer having to know about the setting: references and bindings stay exactly as they are, the capture just stops.
- **`AudioCollector` can restart after a PipeWire error.** `std::jthread` stays joinable once its function returns, so the `joinable()` check in `start()` meant a single `PW_STREAM_STATE_ERROR` - exactly what another client taking the monitor or the default sink disappearing causes - left the capture dead for the rest of the session. Restarts are budgeted (3, 2s apart, reset after 30s of healthy capture) so a capture that cannot work never turns into a reconnect loop on the daemon.
- **No more phantom device-change toasts.** The default node object is replaced whenever the graph moves, and its `description` arrives after the object does, so reacting to the object announced a change - sometimes as "Now using: Unknown Device" - every time something like OBS started. It now tracks the device's own name and says nothing until that is known, including at startup where the placeholder used to seed a change on the first real name.

## How did you test it?

Built the plugin clean, then drove the provider standalone with `qml` while counting sink-monitor captures in `pw-dump`:

| case | capture nodes |
|---|---|
| baseline (shell only) | 2 |
| `enabled: true` + `ServiceRef` | 3 |
| `enabled: false` + `ServiceRef` | 2 |
| toggled true -> false at runtime | 3 -> 2 |
| after the test app exits | 2 |

So the switch really does release the stream, at startup and at runtime, and the reference counting still tears it down cleanly.

`check_qml_syntax.py`, `check_qml_conventions.py` (no new violations), `check_qml_imports.py`, `check_config_references.py` and `test_repo_integrity.py` all pass.

## Type

- [x] Bug fix

## Notes for reviewers

- I could not reproduce the Vesktop freeze itself here, so I have not claimed this closes #402 - it removes the shell from the audio graph on demand, fixes the toasts @Lycham reported, and stops a stream error from silently killing the capture. If a reporter confirms the toggle clears it, that also confirms the audio capture is the remaining half.
- The retry budget is deliberately small. An unbounded reconnect is the failure mode that would hurt every other client on the daemon, which is the opposite of what this issue needs.
- `services.audioCapture` is on by default, so nothing changes for anyone not hitting this.

### DIFF
--- a/shell/modules/nexus/pages/ServicesPage.qml
+++ b/shell/modules/nexus/pages/ServicesPage.qml
@@ -218,6 +218,13 @@ PageBase {
             onClicked: root.nState.openSubPage(1)
         }
 
+        ToggleRow {
+            text: qsTr("Audio capture")
+            subtext: qsTr("Needed by the visualisers. Turn off if screen sharing or a camera stalls while the shell is running")
+            checked: GlobalConfig.services.audioCapture
+            onToggled: GlobalConfig.services.audioCapture = checked
+        }
+
         StepperRow {
             label: qsTr("Visualiser bars")
             subtext: qsTr("Number of bars in the audio visualisers")
--- a/shell/plugin/src/Caelestia/Config/serviceconfig.hpp
+++ b/shell/plugin/src/Caelestia/Config/serviceconfig.hpp
@@ -26,6 +26,13 @@ class ServiceConfig : public ConfigObject {
         bool, useTwelveHourClock, QLocale().timeFormat(QLocale::ShortFormat).toLower().contains(u"a"_s))
     CONFIG_GLOBAL_PROPERTY(QString, gpuType)
     CONFIG_GLOBAL_PROPERTY(int, visualiserBars, 60)
+    // The visualisers and the beat tracker feed off a PipeWire capture of the
+    // sink monitor, which the shell otherwise holds open for as long as
+    // something is drawing them. Some setups cannot share that capture: the
+    // screen shares and camera captures in issue #402 stall while the shell
+    // holds one. Turning this off keeps the shell out of the audio graph
+    // entirely, at the cost of the visualisers.
+    CONFIG_GLOBAL_PROPERTY(bool, audioCapture, true)
     CONFIG_GLOBAL_PROPERTY(qreal, audioIncrement, 0.1)
     CONFIG_GLOBAL_PROPERTY(qreal, brightnessIncrement, 0.1)
     CONFIG_GLOBAL_PROPERTY(qreal, maxVolume, 1.0)
--- a/shell/plugin/src/Caelestia/Services/audiocollector.cpp
+++ b/shell/plugin/src/Caelestia/Services/audiocollector.cpp
@@ -8,6 +8,8 @@
 #include <spa/param/audio/format-utils.h>
 #include <spa/param/latency-utils.h>
 #include <stop_token>
+#include <qelapsedtimer.h>
+#include <qtimer.h>
 #include <vector>
 
 Q_LOGGING_CATEGORY(lcAc, "caelestia.services.ac", QtInfoMsg)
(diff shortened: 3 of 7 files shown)


WinTone01:
Test result from a reporter's machine, and it is a negative one worth recording: **this does not fix #402.**

With `services.audioCapture` off the shell holds no capture stream at all - verified in `pw-dump`, the `stream.capture.sink=true` node is gone and only the idle sound-effect output remains - and Vesktop still fails exactly as before. So the shell's audio capture can be crossed off the list of causes.

What the same session turned up, in case it helps whoever picks this up next:

- Vesktop's **renderer process segfaults**, it does not merely freeze. Four `SIGSEGV` coredumps in one session (`coredumpctl`, `/usr/lib/vesktop/vesktop --type=renderer`, thread `ThreadPoolForeg`). The binary is stripped and the cores carry no library mappings, so no usable backtrace - but it is a native crash, not a hang, which points at a native module rather than at the web app.
- Vesktop 1.6.7 bundles **venmic** (`@vencord/venmic` appears in `app.asar`), which is the part that builds the virtual sink and relinks streams for audio sharing.
- Earlier in the same session, before an unrelated restart: `xdg-desktop-portal-kde: PipeWire remote error: -2 target not found`, `Stream error: target not found`, `pipewire-pulse: [discord_capture] timeout on stream`, and `wireplumber: link failed: 1 of 1 PipeWire links failed to activate`. So the portal's screencast stream lost its target node and Discord's capture stream timed out - consistent with @Lycham's "OBS can't catch my screen".
- `bar.livePreviews` was still on for that test, so the screencast half is not eliminated yet. That is the next variable to isolate.

The PR still stands on its own - the phantom "Now using: Unknown Device" toasts @Lycham reported are fixed, `AudioCollector` no longer dies permanently on a single PipeWire error, and the switch is there for anyone who wants the shell out of the audio graph - but I have retitled expectations: it is a cleanup that rules out one suspect, not the fix for #402.

WinTone01:
Closing this. Testing on a reporter's machine ruled the audio capture out as the cause of #402 - with `services.audioCapture` off the shell holds no capture stream at all (verified in `pw-dump`) and the failure is unchanged - so the premise this was built on does not hold, and I would rather not leave a PR open that looks like a fix for that issue.

Recording where the investigation actually landed, since it points somewhere else entirely:

Vesktop launched from a terminal does **not** fail. The same Vesktop launched from the shell does. And the coredumps agree: the crashed processes are in

```
Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/app-caelestiashell@autostart.service
     User Unit: app-caelestiashell@autostart.service
```

so they were spawned as children of the shell and stayed in its unit. `Apps.launch()` calls `DesktopEntry.execute()` and the dock calls `execDetached` directly, so a launched app inherits the shell's cgroup and its whole environment - which, because the autostart wrapper runs `exec stdbuf -oL -eL quickshell`, includes `LD_PRELOAD=/usr/lib/coreutils/libstdbuf.so`, `_STDBUF_O/E`, `QSG_RENDER_LOOP=threaded`, `QS_*` and `QML2_IMPORT_PATH`.

`services.useSystemd` (app2unit) already exists for exactly this, but it defaults to off and only covers the dock, not the launcher.

I have not confirmed which part of that inherited context is fatal - a headless Chromium with the same `LD_PRELOAD` starts fine here - so I am not opening a replacement PR until the reporter's A/B narrows it down. Wanted the trail on record either way.

PR #586, opened 2026-09-02
fix: stop handing launched applications a stdout that goes nowhere

### LINKED ISSUE #402: Error that causes Vesktop (discord) crash when using camera and/or screen share with audio

WinTone01:
Fixes #402.

## What does this change?

Applications launched from the shell inherit its stdio, and the shell is started with `quickshell --daemonize`, which detaches from the terminal by pointing stdout and stderr at `/dev/null`. So every application started from the launcher, the dock or the context menu runs with both descriptors on `/dev/null`.

Vesktop deadlocks in exactly that state as soon as a call starts. That is the whole of #402: "it freezes when I share with audio or turn on my camera", "OBS can't catch my screen", "it started once I downloaded caelestia", and the randomness - it depends on how the app happened to be launched.

It reproduces with no shell involved at all, 3/3 each way on the reporter's machine:

```
vesktop >/dev/null 2>&1     # deadlocks on joining a call
vesktop                     # fine
vesktop </dev/null          # fine
vesktop 2>&1 | cat          # fine
```

So it is the descriptors themselves - not the terminal, not buffering, not stdin.

**The fix:** stop daemonizing. Nothing that starts the shell actually needed it.

- `scripts/10-autostart.sh` - the autostart entry already runs under a systemd user unit that supervises the process and puts its stdio on the journal. `--daemonize` was throwing that away.
- `src/bin/caelestia-shell-ipc` and `update.sh` - start the shell as a transient user service instead. That returns immediately and survives the terminal the same way `-d` did, with stdio on the journal. Both keep the old path as a fallback when `systemd-run` is unavailable.

Dropping `-d` also retires the `stdbuf` wrapper. Journald stdio is what that line-buffering hack was working around - it could never have worked while the destination was `/dev/null` - and it leaked `LD_PRELOAD=/usr/lib/coreutils/libstdbuf.so` into every launched application on top.

## How did you test it?

Bisected on a machine that reproduces #402, ruling out one variable at a time by reproducing each from a terminal:

| suspect | test | result |
|---|---|---|
| the shell's audio capture | `services.audioCapture` off, verified in `pw-dump` | still broken |
| inherited environment | `env LD_PRELOAD=... QSG_RENDER_LOOP=... vesktop` | fine |
| inherited GPU render node fd | `bash -c 'exec 9<>/dev/dri/renderD128; vesktop'` | fine |
| fd limits | soft `RLIMIT_NOFILE` is 1024 on both sides | not a difference |
| cgroup / systemd unit | moved the process into the shell's cgroup, then launched | fine |
| ignored SIGINT/SIGQUIT | `bash -c 'trap "" INT QUIT; exec vesktop'` | fine |
| 10 inherited eventfds | `python3 -c '[os.eventfd(0, flags=0) ...]; os.execvp(...)'` | fine |
| **stdio on /dev/null** | `vesktop >/dev/null 2>&1` | **deadlocks, 3/3** |

After the change, on the same machine:

```
$ readlink /proc/$(pgrep -f "quickshell.*caelestia/shell.qml")/fd/1
socket:[62519940]
```

journald instead of `/dev/null`, and Vesktop no longer freezes when launched from the shell - camera and audio share both work.

## Type

- [x] Bug fix

## Notes for reviewers

- Why Vesktop deadlocks on `/dev/null` specifically - a pipe is fine, a tty is fine - is a Vesktop/venmic question, not ours. What is clearly ours is that a desktop shell should not impose its own stdio on the applications it launches.
- The transient-service change to `caelestia-shell-ipc` is the part worth a second look. `systemd-run --user --unit=caelestia-shell` gives the same "returns immediately, outlives the terminal" behaviour as `-d`, and `_is_running` still finds the process the same way.
- `update.sh` now prefers the IPC wrapper over `caelestia shell -d`, because the CLI daemonizes and would put the stdio back on `/dev/null`.


WinTone01:
Pushed a second commit after the reporter pointed out a related case this did not cover.

MangoHud users start the shell as

```
env -u MANGOHUD -u MANGOHUD_CONFIG -u LD_PRELOAD caelestia shell -d
```

so the overlay does not land on the shell itself. Every application launched from the shell then inherits that stripped environment and loses the overlay too. Same root cause as the stdio problem: a launched application should not be running in the shell's process context.

That command also shows the gap in the first commit. Dropping `--daemonize` from the autostart entry fixes the shell we start, but `caelestia shell -d` in the CLI still daemonizes, so anyone starting it that way keeps `/dev/null` stdio and #402 with it. The launch path should not depend on how the shell was started.

So launches now go through the systemd user manager, in `utils/Launch.qml` - `app2unit` where available (it names the unit after the desktop entry), `systemd-run` otherwise. That gives an application:

- the **session** environment rather than the shell's. Verified: a variable unset in the caller but set in the session reaches the application.
- **journald stdio**, whatever the shell's own stdio happens to be.
- its **own unit**, so it does not die with the shell or share its `OOMPolicy=stop`.

The launcher, the dock, the dock popout and the desktop context menu all go through it. The launcher previously called `DesktopEntry.execute()`, which spawns directly.

`services.useSystemd` already existed for the dock's app2unit path and now covers every launch, so it defaults to on; setting it to false restores the old behaviour for anyone who wants it.

WinTone01:
**1. `restart_shell.sh`** - yes, and it was worse than the other entry points. It ended with `caelestia shell -d`, and all three call sites (AppearancePage, PluginsPage, the Utilities toggle) run it as `nohup ... >/dev/null 2>&1 & disown` - so even without `-d` the shell would have come back on `/dev/null`. A Restart Shell from the UI undid what login now gets right, until the next login.

Fixed in 03d855b: it starts a transient user service like the other entry points, delegating to `caelestia-shell-ipc start` when present. systemd gives the unit journald stdio regardless of what the script itself was started with - verified:

```
$ bash -c 'exec systemd-run --user --quiet --collect --unit=t -- sleep 6 >/dev/null 2>&1 </dev/null'
fd1 = socket:[559470]
```

It also resets a lingering `caelestia-shell.service` first, so a restart after a crash does not fail on the unit name.

**2. krunner / kstart vs app2unit** - worth separating the two halves of this PR.

For *starting the shell* neither applies: that is a session service, not an application launch, so it uses `systemd-run --user` directly.

For *launching applications* (the `Launch` helper) the KDE-native equivalent is `kstart` - krunner is the search UI and delegates to KIO's `ApplicationLauncherJob`, which is what `kstart` exposes on the command line. I measured it against `systemd-run` from a `/dev/null` caller:

| | stdio | own unit | environment |
|---|---|---|---|
| `kstart` | journald ✅ | `app-/usr/bin/sleep@<rand>.service` ✅ | **inherits the caller's** ❌ |
| `systemd-run` / `app2unit` | journald ✅ | ✅ | from the session ✅ |

```
$ bash -c 'MARKER=fromcaller kstart /usr/bin/sleep 8 ...'        -> MARKER=fromcaller
$ bash -c 'MARKER=fromcaller systemd-run --user -- sleep 8 ...'  -> <absent>
```

So kstart fixes the stdio half of #402 and the lifetime half, but not the environment half: an app launched from a shell started as `env -u MANGOHUD -u LD_PRELOAD caelestia shell -d` would still inherit the stripped environment and lose the overlay. app2unit and systemd-run go through the systemd user manager, so the app starts from the session environment instead. app2unit also names the unit after the desktop entry, while kstart with a raw argv names it after the binary path.

That said, kstart's advantage is real: it ships with Plasma, so on a KDE port it is always there, whereas app2unit is an extra package (`Shortcuts.qml` already uses kstart for its app shortcuts). If you would rather lean on the Plasma-native path I am happy to put it in the probe - either ahead of `systemd-run` as the guaranteed-present fallback, or ahead of app2unit if you consider the environment caveat acceptable. Your call, it is a two-line change to `Launch.qml`.

0xSolanaceae (approved):
LGTM!

PR #603, opened 2026-09-04
feat: synchronize translations with Crowdin

WinTone01:
Follow-up to #550. Translating currently means a checkout, Qt Linguist and a pull request, which is a lot to ask of someone who just wants to contribute a language. Crowdin reads Qt `.ts` natively, so catalogues can be worked on in the browser and arrive here as a pull request instead.

`crowdin.yml` maps the catalogues, `.github/workflows/crowdin.yml` runs the synchronization on pushes to `dev` that touch the shell sources, and it can be triggered by hand.

Two things about this repository needed accommodating, and both are worth knowing before changing anything here:

- **The English source catalogue is generated, not committed.** Crowdin needs one to slice into the target languages, but it is several thousand lines of empty translations and does not belong in the repo — it is what we removed during review of #550. The workflow rebuilds it from the QML immediately before uploading and never commits it.
- **A Crowdin pull request carries `.ts` only.** Since a build without Linguist tools installs the committed `.qm` as it stands, a `.ts` arriving without its `.qm` would quietly ship the previous translation. The workflow recompiles them on the localization branch so the pair always moves together.

The file mapping uses `%locale_with_underscore%` rather than `%two_letters_code%`, so a language needing a region (`pt_BR`, `zh_CN`) lands on the name Qt expects. For a plain language they are identical, so `tr` stays `caelestia_tr.ts`.

**This needs two things from you before it can run**, and I cannot do either:

1. A Crowdin project. It is free for open source through their OSS programme.
2. Two repository secrets from an account with at least Manager rights on that project: `CROWDIN_PROJECT_ID` and `CROWDIN_PERSONAL_TOKEN`.

Until those exist the workflow is inert — it is gated on `github.repository`, so it does not run on forks and will not fail anyone's PR.

**On verification**, so it is not assumed: `actionlint` passes with the same invocation and ignores the validate workflow uses, the YAML parses, and the local CI checks pass. I could not exercise the workflow end to end, because that needs the Crowdin project and secrets above. The parts I could check independently are that Crowdin supports Qt TS natively and that `update-translations.sh en` regenerates the source catalogue from the QML — that script path is the one #550 already ships.

Also corrects a line in `translations.md` that says a build without Linguist tools "ships English only". That stopped being true when the compiled catalogues started being committed; it now installs them as they are.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -11,6 +11,7 @@
 [![Debian](https://img.shields.io/badge/Debian-A81D33?logo=debian&logoColor=white&style=flat-square)](https://debian.org)
 [![KDE Plasma](https://img.shields.io/badge/Plasma_6-1D99F3?logo=kde&logoColor=white&style=flat-square)](https://kde.org/plasma-desktop)
 [![License: GPLv3](https://img.shields.io/badge/License-GPLv3-86dbce?style=flat-square)](LICENSE)
+[![Crowdin](https://badges.crowdin.net/caelestia-kde/localized.svg)](https://crowdin.com/project/caelestia-kde)
 
 </div>
 
--- a/.github/docs/translations.md
+++ b/.github/docs/translations.md
@@ -81,9 +81,19 @@ Two details are worth knowing before touching that setup:
   `.qm` would quietly ship the previous translation. The workflow recompiles them
   on the localization branch so the pair always moves together.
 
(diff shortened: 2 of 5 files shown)


0xSolanaceae:
this requires @ladybug-me to implement the crowdln secrets in the repo

ladybug-me (approved):
Ready to merge

PR #606, opened 2026-09-04
fix: seed Crowdin with the committed translations on the first run

WinTone01:
Caught before the first sync rather than after, so #603 as merged would have destroyed the Turkish translation on its first run.

Crowdin starts empty. The workflow uploads sources and downloads translations, so the first run exports every target file untranslated and opens a pull request replacing a finished `caelestia_tr.ts` with a blank one. `upload_translations` defaults to `false` in the action, and that is the right default for every run except the first.

Manual runs now take a `seed_translations` input that uploads the committed catalogues alongside the sources, so the download returns what is already here. It stays off otherwise: translations are edited in Crowdin, and pushing the committed ones back on every run would let a stale local catalogue overwrite newer work there.

**So the first run should be triggered by hand with that box ticked.** After that the workflow runs on pushes to `dev` as before.

One more thing worth knowing when you add the secrets: `CROWDIN_PROJECT_ID` is the numeric project ID from the project settings, not the `caelestia-kde` identifier that appears in the URL and in the badge. Both are documented now.

Verified with `actionlint` using the same flags and ignores the validate workflow uses, and the YAML parses. The sync itself still cannot be exercised from here — it needs the secrets — so the seeding order is read from the action's documented behaviour (sources, then translations, then download) rather than from a run.

### DIFF
--- a/.github/docs/translations.md
+++ b/.github/docs/translations.md
@@ -97,6 +97,21 @@ seeded first: trigger the workflow by hand once with **Also upload the committed
 catalogues** enabled. Every run after that leaves it off, so translations edited
 in Crowdin are not overwritten by whatever happens to be committed here.
 
+Delete the `i18n/crowdin` branch before seeding if one is already there. The
+branch is where the workflow accumulates downloaded translations, and one left
+over from an earlier run carries that run's files into the next pull request.
+
+Only languages someone has actually translated are exported
+(`skip_untranslated_files`). Without it Crowdin writes a file for every target
+language on the project, and a catalogue of nothing but untranslated entries
+still compiles and still appears in the language picker as a language that
+renders in English.
+
+Catalogues are named by the two-letter code, so Turkish is `caelestia_tr.ts`.
+Two variants of one language -- `pt-BR` and `pt-PT`, `zh-CN` and `zh-TW` --
+would both want the bare code; give them explicit names through
+`languages_mapping` in `crowdin.yml` if both are ever translated.
+
 `CROWDIN_PROJECT_ID` is the numeric project ID from Crowdin's project settings,
 not the `caelestia-kde` identifier that appears in the URL and the badge.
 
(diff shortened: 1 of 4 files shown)


WinTone01:
Two things spotted in the Crowdin UI, and only one of them is a problem.

**`caelestia_en.ts` showing under Turkish is correct.** Crowdin's per-language view lists the *source* files with that language's progress against them, so every target language shows the same `caelestia_en.ts`. Turkish sitting at 0% is just the seeding not having happened yet — that is what the manual first run with **Also upload the committed catalogues** is for.

**English at 0% is a project setting.** English (United States) is among the target languages — the first sync exported a `caelestia_en_US.ts`, which is how I can tell. English is the source language, so there is nothing to translate into it and the figure can never move off 0%.

Worth removing rather than ignoring, because it is not only cosmetic: named by the two-letter code its export would be `caelestia_en.ts`, which is the generated source catalogue and therefore ignored — so a download would quietly discard it every run. Removing English under **Settings -> Languages** on the Crowdin project settles both.

Documented in `translations.md` so it does not get re-added later. No workflow change for it: the export naming and the ignore rule are both right as they are, and a guard in CI would be working around a setting that takes one click.

WinTone01:
On authentication — you asked for `caelestia-automation` and pointed at the other PR-creating workflows, and I had missed that when writing #603. It is in now, and matched against them rather than guessed:

`update-submodules.yml` and `version-release.yml` are the two that open pull requests. Both take `actions/create-github-app-token@v3` with `vars.APP_CLIENT_ID` and `secrets.APP_PRIVATE_KEY`, check out with that token, and commit as `github-actions[bot]`. `version-release.yml` additionally scopes the token with `permission-contents: write` and `permission-pull-requests: write`. This workflow now does the same in both jobs.

Both, because the second one mattered too: the sync job was switched over when the first run failed to open a pull request, but the job that recompiles the catalogues was still checking out and pushing with the default token — pushing commits to a branch the app had created. `secrets.GITHUB_TOKEN` is no longer referenced anywhere in this workflow.

This also removes the need to change any repository setting. `GITHUB_TOKEN` is the one that cannot open pull requests unless Actions is permitted to; the app is not subject to that.

Issue #613, opened 2026-09-05
HyprlandState's KDE fallback never wires up workspaces/monitors/layers — a landmine for future code


0x0nYx:
**Severity: Low-Medium — no live user-facing breakage today, but a guaranteed silent trap for the next contributor**

### Issue type

- [x] Bug - something is broken or not working correctly

### What happened?

While tracing how workspace/monitor data flows through the KDE port, I found that `HyprlandState` — the compatibility shim `HyprlandData.qml` sits on top of — only forwards `windowList`/`windowByAddress`/`addresses`/`activeWindow` to the KWin bridge when Hyprland isn't detected. `workspaces`, `workspaceById`, `workspaceIds`, `activeWorkspace`, `layers`, and `monitors` are left permanently empty under KDE.

Root cause: `shell/plugin/src/Caelestia/Services/hyprlandstate.cpp`, constructor: when `$HYPRLAND_INSTANCE_SIGNATURE` is unset it wires `PlasmaWindows::windowAdded`/`handleLost` to `onKWinWindowListChanged()`/`onKWinActiveWindowChanged()` and returns — it never calls `updateWorkspaces()`/`updateMonitors()`/`updateLayers()`, and those functions are hardwired to `makeRequestJson()`, which talks to `m_requestSocket` — a Hyprland IPC socket path that's left as the empty string `""` on the KDE path. Calling any of them under KDE just fails to connect and (for `workspaces`) clears to empty, or (for `monitors`/`layers`/`activeWorkspace`) silently leaves the property stale, since those three don't clear on failure at all. Right now this is latent: I checked every current QML consumer of `HyprlandData` (`Dock.qml`, `GameMode.qml`) and none of them touch `.workspaces`/`.monitors`/`.layers` — they only use the two properties that *are* wired. But the class presents a full Hyprland-shaped API with no runtime indication that six of its ten properties are dead under KDE, so the next widget or community PR that reaches for `HyprlandData.monitors` expecting parity with `windowList` will get silently empty data with no error.

### Steps to reproduce

1. Under Plasma (no `HYPRLAND_INSTANCE_SIGNATURE` set), from a QML console or a quick test binding, read `HyprlandData.monitors`, `.workspaces`, or `.layers`.
2. All three are empty arrays/maps regardless of actual monitor/desktop state, with no warning logged beyond the one-time constructor message about using the KDE bridge.

### Caelestia version

main branch

### Distro

Arch

### Logs or screenshots

[...]

PR #618, opened 2026-09-05
fix(kwin): sync activeOutputName when active window changes

aroaxinping:
## What does this change?

`buildWindowList()` updates `m_activeWindow` on focus change but never
calls `setActiveOutputName()`, so it stays empty the whole session.
`Hypr.focusedMonitor` falls back to index 0 when it's empty — on
multi-monitor this means the focused monitor is always wrong.

Fix: pass the `output` field (already in `windowToVariant()`) to
`setActiveOutputName()` when the active window changes.

## How did you test it?

Two-monitor setup, verified `activeOutputName` updates on focus change.

## Type

- [x] Bug fix

Related to #613

### DIFF
--- a/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
+++ b/shell/plugin/src/Caelestia/Services/kwinactivewindowbridge.cpp
@@ -267,7 +267,14 @@ void KWinActiveWindowBridge::buildWindowList() {
     if (activeWindowFound && m_activeWindow != newActiveWindow) {
         m_activeWindow = newActiveWindow;
         emit activeWindowChanged();
-        
+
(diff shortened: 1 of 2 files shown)


ladybug-me (changes requested):
requires `Q_INVOKABLE void setActiveOutputName(const QString &outputName);` definition in kwinactivewindowbridge.hpp to prevent `  WARN scene: @modules/background/Background.qml[51:-1]: TypeError: Property 'setActiveOutputName' of object caelestia::services::KWinActiveWindowBridge(0x7f3bed92cd70) is not a function`

ladybug-me:
Checks out just one change required

ladybug-me (approved):

PR #619, opened 2026-09-05
fix(discord): make IPC reconnect non-blocking

### LINKED ISSUE #615: DiscordIpc::checkReconnect() runs up to 30 sequential blocking socket connects on the main thread every 5 seconds
**Severity: Medium — periodic UI stutter for any user without Discord running, worse with a stale socket file**

### Issue type

### What happened?

With Discord/Vesktop closed, the shell has a small but very regular hitch roughly every 5 seconds — most noticeable as a stutter in animations or a delayed response to input at that cadence.

Root cause: `shell/plugin/src/Caelestia/Services/discordipc.cpp`, `checkReconnect()` (invoked from `m_reconnectTimer`'s `timeout()`, interval 5000ms, on the object's own thread — the QML engine thread) does:
```cpp
for (int slot = 0; slot <= 9; ++slot) {
    QString pipePath = runtimeDir + "/discord-ipc-" + QString::number(slot);
    m_socket->connectToServer(pipePath);
    if (m_socket->waitForConnected(500))
        return;
}
// then the same 0..9 loop again for each of 2 Flatpak app-ids
```
That's up to 30 sequential `connectToServer()` + `waitForConnected(500)` pairs per invocation, all synchronous and all on the QML thread — `waitForConnected` spins the event loop internally but still blocks the calling function (and therefore blocks whatever queued QML work was waiting behind it) until each attempt resolves or hits its own 500ms timeout. For a nonexistent socket path this usually fails fast (ENOENT), so the common case is a small but real cumulative hitch; but a **stale leftover socket file** from a crashed Discord/Vesktop session (the path exists but nothing is `accept()`-ing on it) turns one of those 30 attempts into a full 500ms hard freeze of the shell, recurring every 5 seconds until the stale file is cleaned up or Discord is relaunched.

### Steps to reproduce

1. Enable the Discord Rich Presence integration, then kill Discord/Vesktop with `kill -9` so it leaves its IPC socket file behind under `$XDG_RUNTIME_DIR/discord-ipc-0` instead of cleaning it up.
2. Watch shell responsiveness — every 5 seconds there's a visible ~500ms freeze while `checkReconnect()` blocks on the stale socket.
3. `ls -la $XDG_RUNTIME_DIR/discord-ipc-*` confirms the stale file is still present and nothing is listening on it (`fuser` shows no owning process).

### Caelestia version

main branch

### Distro

CachyOS

### Logs or screenshots

No error output — `waitForConnected` returning `false` after its timeout isn't logged; the only evidence is the periodic stutter and the leftover socket file.

---
_Note: the 0..9-slot + Flatpak-app-id scan is exactly the fix landed in #332 for #315 (native Discord/Vesktop only trying `discord-ipc-0`, breaking under Flatpak). That fix was correct in scope but turned a single fast connect attempt into up to 30 sequential blocking ones on the QML thread — this issue is the performance regression that fix introduced, not a duplicate of it._

aroaxinping:
## What does this change?

`checkReconnect()` called `waitForConnected(500)` up to 30 times per
invocation — 10 native slots + 2 Flatpak app-ids × 10 slots — all on
the QML thread. With Discord closed, small periodic hitch every 5 s.
With a stale socket file from a crashed Discord/Vesktop, one of those
attempts hard-freezes the shell for 500 ms, every 5 s, until the file
is gone.

Replaces the blocking loop with a signal-driven state machine:
- `checkReconnect()` builds the candidate list and calls `tryNextPath()`
- `tryNextPath()` skips non-existent paths with `QFile::exists()` and
  calls `connectToServer()` without `waitForConnected()`
- `onError()` advances to the next candidate via a queued
  `invokeMethod()` to avoid recursion on synchronous failures

## How did you test it?

Killed Vesktop with `kill -9` to leave a stale socket, confirmed no
freeze. Also verified normal connect/disconnect cycle still works.

## Type

- [x] Bug fix

Closes #615

### DIFF
--- a/shell/plugin/src/Caelestia/Services/discordipc.cpp
+++ b/shell/plugin/src/Caelestia/Services/discordipc.cpp
@@ -18,7 +18,8 @@ enum class Opcode : int32_t {
 };
 
 DiscordIpc::DiscordIpc(QObject* parent)
-    : QObject(parent), m_socket(new QLocalSocket(this)), m_reconnectTimer(new QTimer(this)), m_connected(false)
+    : QObject(parent), m_socket(new QLocalSocket(this)), m_reconnectTimer(new QTimer(this)),
+      m_connectTimeout(new QTimer(this)), m_connected(false)
 {
     connect(m_socket, &QLocalSocket::connected, this, &DiscordIpc::onSocketConnected);
     connect(m_socket, &QLocalSocket::disconnected, this, &DiscordIpc::onSocketDisconnected);
@@ -31,6 +32,15 @@ DiscordIpc::DiscordIpc(QObject* parent)
 
     m_reconnectTimer->setInterval(5000);
     connect(m_reconnectTimer, &QTimer::timeout, this, &DiscordIpc::checkReconnect);
+
+    m_connectTimeout->setSingleShot(true);
+    m_connectTimeout->setInterval(1000);
+    connect(m_connectTimeout, &QTimer::timeout, this, [this]() {
+        if (m_socket->state() == QLocalSocket::ConnectingState) {
+            m_socket->abort();
+            tryNextPath();
+        }
+    });
 }
 
 DiscordIpc::~DiscordIpc() {
@@ -52,6 +62,8 @@ void DiscordIpc::connectIpc(const QString& clientId) {
 
 void DiscordIpc::disconnectIpc() {
     m_reconnectTimer->stop();
+    m_connectTimeout->stop();
+    m_pendingPaths.clear();
     m_clientId.clear();
     m_socket->abort();
     if (m_connected) {
@@ -82,20 +94,27 @@ void DiscordIpc::checkReconnect() {
 }
 
 void DiscordIpc::tryNextPath() {
+    if (m_clientId.isEmpty()) {
+        m_pendingPaths.clear();
+        return;
+    }
+
     while (!m_pendingPaths.isEmpty()) {
         const QString path = m_pendingPaths.takeFirst();
         if (QFile::exists(path)) {
             if (m_socket->state() != QLocalSocket::UnconnectedState)
                 m_socket->abort();
             m_socket->connectToServer(path);
+            m_connectTimeout->start();
             return;
         }
     }
-    // All candidates exhausted — timer retries in 5 s
 }
 
 void DiscordIpc::onSocketConnected() {
-    // Send Handshake
+    m_connectTimeout->stop();
+    m_pendingPaths.clear();
+
     QJsonObject payload;
     payload["v"] = 1;
     payload["client_id"] = m_clientId;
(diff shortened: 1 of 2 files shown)


shell/plugin/src/Caelestia/Services/discordipc.cpp
```diff
@@ -109,6 +113,11 @@ void DiscordIpc::onSocketDisconnected() {
 void DiscordIpc::onError(QLocalSocket::LocalSocketError) {
     emit errorOccurred(m_socket->errorString());
     onSocketDisconnected();
+    // If a reconnect scan is in progress, try the next candidate non-blocking.
+    // Use a queued invoke to avoid re-entering connectToServer() inside its own
+    // error signal, which can recurse on synchronous failures (e.g. ENOENT).
+    if (!m_pendingPaths.isEmpty())
+        QMetaObject::invokeMethod(this, &DiscordIpc::tryNextPath, Qt::QueuedConnection);
```
0xSolanaceae, line 137: m_pendingPaths is only cleared in checkReconnect(). After a successful connect it still has like 30 leftover paths, so this can fire on an unrelated error later - even after the user disables the integration (disconnectIpc() doesn't clear it, and tryNextPath() doesn't check m_clientId). That means a connect can happen after the user turned it off. if you could clear the list on connect/disconnect/disable, or adding a m_clientId check to tryNextPath(), that would be great

shell/plugin/src/Caelestia/Services/discordipc.cpp
```diff
@@ -63,31 +64,34 @@ void DiscordIpc::checkReconnect() {
...
+    tryNextPath();
+}
+
+void DiscordIpc::tryNextPath() {
+    while (!m_pendingPaths.isEmpty()) {
+        const QString path = m_pendingPaths.takeFirst();
+        if (QFile::exists(path)) {
+            if (m_socket->state() != QLocalSocket::UnconnectedState)
+                m_socket->abort();
+            m_socket->connectToServer(path);
```
0xSolanaceae, line 107: the old waitForConnected(500) was the only timeout on a connect attempt. Now if a connect hangs the socket will stay in ConnectingState forever and checkReconnect() bails out on that state, so nothing ever retries - ARPC is stuck until the user toggles the feature off and on. Maybe add a timeout that aborts a stuck connect?

0xSolanaceae (changes requested):
couple additions, thanks!

aroaxinping:
Addressed both points. m_pendingPaths is now cleared on successful connect, in disconnectIpc(), and tryNextPath() bails early if m_clientId was cleared. Added a 1s singleshot connect timeout. I think 500ms would also work but native sockets connect in microseconds so the timeout only really fires on dead sockets or slow Flatpak portals, and QFile::exists() already skips most candidates before they hit connectToServer(). Happy to lower it if you prefer.

0xSolanaceae (approved):
LGTM!

PR #638, opened 2026-09-05
/claim #616: Fix hybrid/multi-GPU monitoring — use max for GPU busy% and temps on dev

### LINKED ISSUE #616: GPU usage and temperature are averaged across every /sys/class/drm/card* device, producing meaningless numbers on hybrid-graphics/multi-GPU systems
**Severity: Medium — silently wrong GPU stats on a fairly common hardware configuration**

### Issue type

### What happened?

On my Ryzen APU + discrete AMD GPU desktop, the GPU usage widget shows roughly half of what either GPU is actually doing under load — running a game that pegs the discrete card at 100% while the iGPU sits idle at 0% displays as ~50% in the bar/Nexus resource widgets.

Root cause: `shell/plugin/src/Caelestia/Services/gpu.cpp`, `readGenericUsage()`:
```cpp
const QStringList paths = QDir("/sys/class/drm").entryList({"card*"}, QDir::Dirs | QDir::NoDotAndDotDot);
qreal sum = 0.0; int count = 0;
for (const QString& card : paths) {
    QFile f(QString("/sys/class/drm/%1/device/gpu_busy_percent").arg(card));
    if (f.open(...)) { sum += f.readAll().trimmed().toDouble(&ok); ++count; }
}
const qreal newPerc = count > 0 ? sum / count / 100.0 : 0.0;
```
This walks *every* `cardN` entry that exposes `gpu_busy_percent` and averages them into one number with no attempt to identify which card is the "real"/active GPU (discrete vs integrated, or primary vs secondary in a multi-GPU box). The type-detection script a few lines up (`kTypeDetectScript`) already treats "any card with `gpu_busy_percent`" as sufficient to classify the system as `GENERIC`, so this isn't a one-GPU-assumption edge case — it's the intended code path for any non-Nvidia system, and it breaks the moment there's more than one DRM card. The temperature path has the identical problem: `readGpuTemperature()` calls `sensorslib::gpuPciAverageTemp()` — the function name says it all.

### Steps to reproduce

1. On a system with an AMD iGPU and AMD/other dGPU both exposing `gpu_busy_percent` under `/sys/class/drm/card*/device/` (common on Ryzen "G"/APU desktops paired with a discrete card, and some AMD Advantage laptops).
2. Load only the discrete GPU (e.g. run a game forced onto it) while the iGPU stays idle.
3. Compare `cat /sys/class/drm/card1/device/gpu_busy_percent` (the loaded card) against Caelestia's GPU usage widget — the widget reads roughly half the real value because it's averaged with the idle card's 0-2%.

### Caelestia version

main branch

### Distro

Arch (Ryzen APU + discrete AMD GPU)

### Logs or screenshots

`for f in /sys/class/drm/card*/device/gpu_busy_percent; do echo "$f: $(cat $f)"; done` next to the widget's displayed percentage shows the widget consistently landing at the arithmetic mean of the per-card values, not the max or the active card's own value.

---
_Note: this is a different code path from #314 (`detectTypeOnce()` misclassifying hybrid Nvidia+other systems as pure Nvidia) — that one is about GPU-type detection; this one is the `GENERIC`-path usage/temperature math itself, which averages across cards regardless of which type won detection. Possibly related to the in-progress #588 ("Redo gpu.cpp"), which doesn't currently mention this specific averaging behavior — flagging here in case it isn't already covered by that rewrite._

jialfaro:
## What does this change?

Fixes incorrect averaging of GPU usage and temperatures on hybrid/multi-GPU systems (Closes #616, recreated against `dev` as requested by @ladybug-me in #633).

- `Gpu::readGenericUsage()` now reports the maximum per-card usage across both Pass 1 (`gpu_busy_percent`) and Pass 2 (Intel iGPU frequency ratio) instead of averaging them.
- `sensorslib::gpuPciAverageTemp()` now returns the maximum primary (or fallback) temperature across PCI GPU devices instead of averaging.

## Screenshots (if UI change)
N/A (backend service monitoring calculation)

## How did you test it?
- Verified per-card sysfs nodes `/sys/class/drm/card*/device/gpu_busy_percent` and Intel frequency ratios `/gt/gt0/rps_cur_freq_mhz` under simulated hybrid GPU loads (100% discrete load + 0% idle integrated load), confirming reporting reflects the active GPU (100%) rather than an artificial arithmetic mean (~50%).
- Verified `sensorslib::gpuPciAverageTemp()` selection logic for primary temperature sensors against fallback sensors.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

Per @ladybug-me's feedback on #633, this PR is cleanly built on top of the latest `dev` branch, integrating with both Pass 1 and Pass 2 detection paths.

Closes #616
/claim #616

### DIFF
--- a/shell/plugin/src/Caelestia/Services/gpu.cpp
+++ b/shell/plugin/src/Caelestia/Services/gpu.cpp
@@ -176,8 +176,7 @@ void Gpu::readGenericUsage() {
         QDir(QStringLiteral("/sys/class/drm"))
             .entryList(QStringList() << QStringLiteral("card*"), QDir::Dirs | QDir::NoDotAndDotDot);
 
-    qreal sum = 0.0;
-    int count = 0;
+    qreal maxPerc = -1.0;
 
     // Pass 1: amdgpu (and some newer Xe) cards expose a direct busy percent.
     for (const QString& card : cards) {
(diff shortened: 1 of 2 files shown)


ladybug-me:
Checks out

jialfaro:
“Quick question: does this task come with any kind of reward?”

0xSolanaceae:
@jialfaro the satisfaction of knowing you fixed our math is reward enough

PR #683, opened 2026-09-08
fix: prune the deleted submodules from local git

### LINKED ISSUE #666: update.sh's post-restart "wait for scheme.json" loop runs before the shell is restarted — no-ops or wastes 15s
**Issue type:** Bug - something is broken or not working correctly

**What happened?**

Near the end of `update.sh`, after killing the running shell and before starting the new one:

```bash
# Kill the running shell...
if "$CAELESTIA_BIN" shell -k 2>/dev/null; then
    :
elif [[ -n "$SHELL_IPC" ]] && "$SHELL_IPC" quit 2>/dev/null; then
    :
else
    pkill -f "quickshell.*caelestia/shell.qml" 2>/dev/null || true
fi

STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/caelestia"
SCHEME_FILE="$STATE_DIR/scheme.json"
i=0
while [[ $i -lt 15 && ! -s "$SCHEME_FILE" ]]; do
    sleep 1
    i=$((i + 1))
done

# Start the shell...
if [[ -n "$SHELL_IPC" ]]; then
    "$SHELL_IPC" start 2>/dev/null &
...
```

`scheme.json` is written by the running shell instance (`Colours.qml`/`schemeloader.cpp` persist the applied color scheme there so it can be read pre-auth by the lock screen and other out-of-process consumers). This wait loop sits between the kill and the (re)start — i.e. it's polling for a file to be (re)written by a process that, at the point this loop runs, has just been killed and has not yet been started again.

In practice this loop can never do the thing its position suggests it's for (waiting for the new shell instance to finish writing its color scheme before considering the restart complete):
- If `scheme.json` already exists with non-empty content from the just-killed instance (the overwhelmingly common case — a running shell has almost certainly already written it at least once), the `-s` check passes on the very first iteration and the loop is a complete no-op.
- If, for whatever reason, it doesn't exist yet (first-ever run, or the file happens to be mid-atomic-replace at the exact moment the old process died), nothing in this window is ever going to write it — the new shell process hasn't started yet — so the loop just burns the full 15 seconds and falls through regardless.

Either way the loop doesn't do useful synchronization; it's dead weight in the update path, and in the (rarer) failure case it adds a flat, unexplained 15-second stall to every update run that hits it.

**Expected:** If the intent is to confirm the new shell instance has come up and written a fresh color scheme before declaring the restart done, the wait should happen after the `"$SHELL_IPC" start &` (or equivalent) call, not before it.

**Actual:** The wait is positioned before the new shell is started, so it can only ever pass trivially (stale file already present) or stall uselessly for 15s (file absent, nothing running yet to produce it).

**Steps to reproduce**
1. Read `update.sh` around the "Kill the running shell" / "Start the shell" comments — the ordering is directly visible in the diff between the two blocks; no special environment is needed to see the logic is misordered.
2. To observe the 15s stall specifically: `rm -f ~/.local/state/caelestia/scheme.json` immediately after the shell is killed but before `update.sh` reaches the start block (tight timing window, but reliably reproducible by adding a `sleep` locally around the kill step while testing), then watch the update pause for exactly 15 seconds before starting the shell.

**Root cause (file/function level):** `update.sh` — the `while [[ $i -lt 15 && ! -s "$SCHEME_FILE" ]]` loop is placed between the "kill the running shell" block and the "start the shell" block; it should be placed after the start block if its purpose is to confirm the new instance came up.

**Suggested fix:** Move the wait loop to after the `"$SHELL_IPC" start`/`systemd-run`/fallback block, so it's actually polling for output from the instance that was just started.

**Caelestia version:** v2.4.0
**Distro:** Any — pure shell-script ordering issue.

ladybug-me:

> [!WARNING]

## Type

- [ ] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other


scripts/02a-submodules.sh
```diff
@@ -8,6 +8,16 @@ source "$(dirname "${BASH_SOURCE[0]}")/lib/log.sh"
 # Initialize submodules
 if [[ -f "$BUNDLE_DIR/.gitmodules" ]]; then
     info "Initializing submodules..."
-    git submodule sync --recursive >/dev/null 2>&1 || true
-    git submodule update --init --recursive --depth 1 --jobs "$(nproc 2>/dev/null || echo 1)" >/dev/null 2>&1 || warn "Failed to initialize all submodules."
+    # Prune any submodule configured locally that was removed from .gitmodules
+    while IFS= read -r -d '' key; do
```
0xSolanaceae: Identical prune block is added in three more places: scripts/08-build-shell.sh:175,
src/bin/caelestia-update:209 and update.sh:99. please simplify to 1

The repo already has the right home for this: scripts/lib/ (log.sh, privileges.sh).
Suggest scripts/lib/submodules.sh exposing prune_removed_submodules "$dir", sourced
by all four callers, with only the -C target differing.

scripts/02a-submodules.sh
```diff
@@ -8,6 +8,16 @@ source "$(dirname "${BASH_SOURCE[0]}")/lib/log.sh"
...
+        submod="${key#submodule.}"
+        submod="${submod%.url}"
+        if ! git -C "$BUNDLE_DIR" config --file .gitmodules --get "submodule.${submod}.url" >/dev/null 2>&1; then
+            git -C "$BUNDLE_DIR" submodule deinit -f "$submod" >/dev/null 2>&1 || true
```
0xSolanaceae: Verify this path was tested. Once a path is gone from .gitmodules, `git submodule
deinit` typically fails with "No submodule mapping found in .gitmodules for path
'<path>'", and the `|| true` swallows that. So the checkout under the worktree is
never removed, and only .git/modules/<name> gets deleted. Is the leftover directory
expected, or should the worktree copy be removed too?

update.sh
```diff
@@ -156,66 +166,12 @@ info "System tweaks (OSD, desktops, CLI patches) have been re-applied to keep KD
...
-    : # CLI succeeded
-elif [[ -n "$SHELL_IPC" ]] && "$SHELL_IPC" quit 2>/dev/null; then
-    : # IPC wrapper succeeded
+if [[ -x "$RESTART_SCRIPT" ]]; then
+    bash "$RESTART_SCRIPT"
+    echo "Shell restarted successfully!"
 else
-    pkill -f "quickshell.*caelestia/shell.qml" 2>/dev/null || true
+    warn "Restart script not found. Please restart the shell manually."
 fi
```
0xSolanaceae, line 187: Issue #666 asked for this wait to be moved, not deleted: "the wait should happen
after the `"$SHELL_IPC" start &` (or equivalent) call, not before it". Here the loop
is removed and nothing replaces it, so nothing confirms the new instance came up
before the restart is declared done. The stall is gone, but the stated intent is
dropped. If that is deliberate, please say so in the PR body rather than closing
#666 via this PR.

update.sh
```diff
@@ -156,66 +166,12 @@ info "System tweaks (OSD, desktops, CLI patches) have been re-applied to keep KD
...
 
-STATE_DIR="${XDG_STATE_HOME:-$HOME/.local/state}/caelestia"
-SCHEME_FILE="$STATE_DIR/scheme.json"
-i=0
-while [[ $i -lt 15 && ! -s "$SCHEME_FILE" ]]; do
-    sleep 1
-    i=$((i + 1))
-done
-
-# Start the shell. The IPC wrapper is preferred over the CLI here because it
```
0xSolanaceae, line 191: This drops the IPC-wrapper / systemd-run start chain along with its rationale
(issue #402: the CLI's `shell -d` daemonizes, children inherit a dead stdout, and
Vesktop deadlocks). restart_shell.sh instead runs `systemctl --user restart
app-caelestiashell@autostart.service`, which is a different mechanism that only
works when that generated unit exists. If that is intentional, carry the #402
rationale across to restart_shell.sh so it is not silently lost.

0xSolanaceae (changes requested):

update.sh
ladybug-me: @0xSolanaceae The stated intent is now kept but first priority is still using `restart_shell.sh` to maintain a single source of truth for restarting the shell, rather then manual restarting steps which may not contain upto date fixes for starting the shell.

0xSolanaceae (approved):
LGTM!

PR #697, opened 2026-09-10
feat: add GIF recording option and fix drag-reorder bug

Alxkcx:
### 1. GIF Recording (new feature, off by default)
Adds a "Record GIF" option to the screen recorder dropdown menu, gated behind a config toggle per the contributing guidelines.

- `utilitiesconfig.hpp`: new `showGifRecorder` property, **default `false`**
- `UtilitiesPanelPage.qml`: new "GIF Recorder" toggle in Settings -> Utilities panel
- `Record.qml`: new `Record GIF` menu item, only shown when `showGifRecorder` is enabled
- `Recorder.qml`: new `startGif()` that passes `--gif` to the recorder
- `caelestia-record`: new `--gif` flag. Records mp4, then on stop auto-converts to GIF via ffmpeg (720p, 30fps, bayer dither) and removes the source mp4. GIFs are written next to the mp4s in `$XDG_VIDEOS_DIR/Recordings/` so `RecordingList.qml` can find them. A stale GIF marker is cleared on failed starts and on the toggle-stop path so a later plain recording is never converted by accident.
- `RecordingList.qml`: include `recording_*.gif` so GIF sessions show up in the Recordings list
- `caelestia_zh_CN.ts` / `.qm`: Simplified Chinese translations for the three new strings

### 2. Fix drag-reorder bug (BarComponents.qml)
Clicking an item in "Toggle & Rearrange" set `isGlobalDragging` prematurely (on press, before `drag.active`), which made the item fill the entire settings panel.

Fix:
- Only set `isGlobalDragging = true` from `onPositionChanged` once `drag.active` is confirmed.
- Replace `anchors.fill: parent` on `activeDelegate` with explicit `width`/`height`. Note: the important part is dropping the **anchor-driven position**, not the size — `anchors.fill` also anchored the delegate to the top-left of whatever parent it lands in. `width: parent.width` would still rebind to `contentItem` after the `ParentChange` reparent, so it is bound to `delegateWrapper` instead.
- Add `onCanceled` to reset `isGlobalDragging` if the drag is cancelled (Esc / pointer leaves window) rather than released.


src/bin/caelestia-record
```diff
@@ -241,8 +260,33 @@ case "$ACTION" in
 
     stop)
         if _is_recording; then
+            # Capture recording path before _stop_recording clears it
+            local saved_path=""
```
0xSolanaceae: local only works inside a function, and this case block is at the top level of the script, so bash exits with local: can only be used in a function. Because we run set -euo pipefail, that kills the whole --stop path before _stop_recording runs, so nothing gets stopped or finalized - including plain mp4 recordings. Please just use saved_path="".

src/bin/caelestia-record
```diff
@@ -182,6 +191,16 @@ case "$ACTION" in
...
         [[ "$rr" -le 0 ]] && rr=60
 
+        # GIF mode: mark this session so stop will auto-convert
+        if [[ "$GIF_MODE" == "true" ]]; then
+            if ! command -v ffmpeg >/dev/null 2>&1; then
+                _notify critical "GIF failed" "ffmpeg not found, cannot convert to GIF"
+                rm -f "$LOCK_FILE"
+                exit 1
+            fi
+            touch "$GIF_MARKER"
```
0xSolanaceae: The marker is written here before the recorder is confirmed started, but the only places it gets removed are inside the stop) GIF branch (lines 271 and 287). If the recording fails to start (line 241-ish "Recording failed" exit 1), or if a recording is stopped through the toggle path at line 175, the marker stays behind. The next plain recording then gets converted to a GIF and its mp4 deleted without the user asking. please clear the marker on those paths too, or better, store the mode next to the current recording path so it can't outlive the session?

src/bin/caelestia-record
```diff
@@ -241,8 +260,33 @@ case "$ACTION" in
...
+
             _stop_recording
-            _notify normal "Recording stopped" "Recording saved to $RECS_DIR"
+
+            # GIF conversion: if marker exists and we have a valid recording, convert it
+            if [[ -f "$GIF_MARKER" ]] && [[ -n "$saved_path" ]] && [[ -f "$saved_path" ]]; then
+                rm -f "$GIF_MARKER"
+                _notify normal "GIF" "Converting to GIF..."
+
+                gif_dir="${RECS_DIR}/gif"
```
0xSolanaceae: The GIF lands in a gif/ subfolder, but the only place the shell lists recordings is RecordingList.qml, which is a non-recursive FolderListModel filtered to recording_*.mp4. So a GIF session leaves the Recordings list empty and there's no way to find the file from the UI (_open_dir opens the parent folder). Worth either writing GIFs next to the mp4s or teaching that list about them.

shell/modules/utilities/cards/Record.qml
```diff
@@ -116,6 +116,15 @@ StyledRect {
                             Recorder.start(["-s","-i"]);
                         }
                     },
+                    MenuItem {
```
0xSolanaceae, line 119: please default the menu config toggle to false

shell/modules/nexus/pages/panels/taskbar/BarComponents.qml
```diff
@@ -104,7 +104,8 @@ PageBase {
             ConnectedRect {
                 id: activeDelegate
 
-                anchors.fill: parent
+                width: parent.width
```
0xSolanaceae: Good catch on the premature flag. One thing to double-check: the delegate is reparented to root.flickable.contentItem by the ParentChange at line 122, and width: parent.width is still a binding, so it rebinds to the new parent on reparent just like anchors.fill did. Is it the loss of the anchor-driven position (rather than the size) that fixes the "fills the panel" symptom? Worth a line in the description so it doesn't get reverted later.

shell/modules/nexus/pages/panels/taskbar/BarComponents.qml
```diff
@@ -135,12 +136,17 @@ PageBase {
                     enabled: !isPlaceholder
 
                     onPressed: {
-                        root.isGlobalDragging = true;
                         root.globalDragSourceList = sourceList;
                         root.globalDragSourceIndex = delegateWrapper.index;
                         root.globalDragHoveredList = sourceList;
                     }
 
+                    onPositionChanged: {
```
0xSolanaceae, line 144: If the drag gets cancelled rather than released (Esc, or the pointer leaves the window), does onReleased still run? If not, isGlobalDragging could stay stuck true and leave the drop targets highlighted.

0xSolanaceae (changes requested):
great work, thanks for the features! re-request when comments have been addressed

shell/modules/utilities/cards/Record.qml
0xSolanaceae: this has still not been resolved

0xSolanaceae (changes requested):

shell/modules/utilities/cards/Record.qml
Alxkcx: Added a `showGifRecorder` config property (default `false`) in `utilitiesconfig.hpp`, a toggle in the Utilities settings page, and bound the menu item visibility to it. GIF recording is now off by default, matching the contributing guideline.

0xSolanaceae (approved):
thanks!

PR #698, opened 2026-09-10
fix:update about page to display actual plugin count

LuYishan-4:
> [!WARNING]

## What does this change?

Add plugin count support to the Caelestia shell and About page.

The plugin IPC handler is registered by `PluginLoader.qml`, keeping it alongside
the plugin-loading logic instead of placing it directly in `shell.qml`.

The About page displays the number of enabled plugins through:

```bash
caelestia shell plugins count
```

The count includes enabled Quickshell plugins and enabled KWin effects, while
disabled plugins are excluded.

## How did you test it?

- Updated the active Quickshell configuration under:
  ```text
  ~/.config/quickshell/caelestia/
  ```
- Restarted the Caelestia shell.
- Verified the IPC target:
  ```bash
  caelestia-shell-ipc show
  ```
- Verified the plugin count:
  ```bash
  caelestia shell plugins count
  ```
- Confirmed the count includes enabled Quickshell plugins and KWin effects.
- Confirmed disabled plugins are excluded from the count.

## Type

- [x] Bug fix
- [x] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

The `plugins` IPC target is owned by `PluginLoader.qml`. The count is based on
discovered plugin metadata and each plugin's `enabled` state. KWin effects are
included alongside Quickshell plugins, while unrelated or disabled entries are
excluded.


shell/services/PluginLoader.qml
```diff
@@ -25,6 +27,16 @@ Item {
 
     signal pluginsReloaded()
 
+    function updateEnabledCount() {
```
0xSolanaceae, line 34: check_qml_conventions.py reports a new [section-order] violation here because the IpcHandler is a child object but is declared above the functions. Move it to the end of the file with the other children

shell/services/PluginLoader.qml
```diff
@@ -281,6 +306,7 @@ Item {
                     if (meta.enabled) {
                         pluginLoader.loadPlugin(meta);
                     }
+                    pluginLoader.updateEnabledCount();
```
0xSolanaceae, line 322: This call can't change anything. updateEnabledCount() recounts from discovered, but addPluginToAvailable only appends to CaelestiaApi.plugins.available (line 309), so discovered is untouched. Nothing recounts after removePluginFromAvailable (line 238) either, so uninstalling leaves the count too high. It also counts every enabled entry regardless of type, so a theme would be included.

shell/modules/nexus/pages/AboutPage.qml
```diff
@@ -151,7 +158,7 @@ PageBase {
             first: true
             last: true
             label: qsTr("Loaded plugins")
-            value: root.pluginCount.toString()
+            value: root.pluginCount || "…"
```
0xSolanaceae, line 161: The row is labelled "Loaded plugins" (line 160) but this is the enabled count, and it includes KWin effects that loadPlugin never loads (that function returns early unless type === "quickshell"). PluginLoader.loadedCount is maintained at lines 14, 54, 75 and 131 and is read nowhere. Use that value here, or rename the label to match.

0xSolanaceae (changes requested):
good work, thanks.
7228f427 only adds a trailing newline to TROUBLESHOOTING.md and "Docs" isn't ticked in the Type list - tick it or drop the commit

shell/services/PluginLoader.qml
```diff
@@ -25,6 +27,19 @@ Item {
 
     signal pluginsReloaded()
 
+    function isCountedPlugin(meta) {
+        return meta.enabled && (meta.type === "quickshell" || meta.type === "kwin");
```
0xSolanaceae: The type check uses "kwin", but the manifest enum is quickshell, kwineffect and theme - see plugin.schema.json and ingestion-contract.md in caelestia-kde-plugins, and the plugin PR template. Nothing emits "kwin", so kwineffect entries never match and KWin effects are excluded, which is the opposite of what the description says. Should this be meta.type === "kwineffect"?

0xSolanaceae (changes requested):
one more fix, thanks!

0xSolanaceae (approved):
thanks!

PR #699, opened 2026-09-10
feat: add SDDM theme as optional installer module

### LINKED ISSUE #632: feat: Add SDDM theme as optional installer module
## Summary

I built an SDDM theme for Caelestia ([caelestia-sddm](https://github.com/aroaxinping/caelestia-sddm)) and want to add it as an optional toggle in the installer.

## What it does

Two theme variants, full and mini. The full one has a clock, system info, tech quotes, and avatar. The mini one is a centered login card.

Wallpaper, colors, and avatar sync automatically through the Caelestia CLI posthook. Video wallpapers work too. Qt6/QML, runs on Wayland.

## How the integration would work

A new toggle `INSTALL_SDDM` in `menu.json` under Appearance, off by default. A new script `05-sddm-theme.sh` that clones the repo and installs the theme. An entry in `Runner.cpp` to run the step.

The theme repo stays separate, the dots installer just points to it.

PR is ready, will open when this gets the green light.

aroaxinping:
Material You themed SDDM login screen. The Caelestia lockscreen covers KDE's lock but not the display manager, so a reboot or logout shows the default SDDM greeter. This adds a matching login screen with wallpaper and color sync.

Theme source is included under src/sddm/ (full and mini variants, sync script, templates). The install script copies the selected theme to /usr/share/sddm/themes/caelestia, sets up the posthook for auto-sync, and creates a sudoers drop-in for passwordless sync.

Toggled with INSTALL_SDDM in Appearance (off by default, on in Full profile).

| Full | Mini |
|:---:|:---:|
|  |  |

**Files changed**

src/sddm/ — full and mini theme source, sync script
scripts/05-sddm-theme.sh — install step
installer/menu.json — INSTALL_SDDM toggle
installer/src/Runner.cpp — step entry and skip logic

Closes #632


aroaxinping:
@0xSolanaceae updated, theme source is now committed under src/sddm/ instead of pulling from an external repo at install time.

src/sddm/themes/full/assets/google-sans-flex/GoogleSansFlex.ttf
0xSolanaceae, line 1: this is duplicate, we already have it in the repo

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,119 @@
...
+
+echo
+echo ""
+info "Installing Caelestia SDDM theme"
+echo ""
+
+THEME_NAME="caelestia"
+INSTALL_DIR="/usr/share/sddm/themes/$THEME_NAME"
+SYNC_SCRIPT="$INSTALL_DIR/scripts/sync.sh"
+THEME_SOURCE="$SRC_DIR/themes/full"
```
0xSolanaceae: I see nothing here that provides the user the choice to switch between full and mini, it looks like you just hardcode full. `themes/mini/` doesn't look like it's touched

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,119 @@
...
+
+if [[ ! -d "$THEME_SOURCE" ]]; then
+    die "SDDM theme source not found at $THEME_SOURCE"
+fi
+
+# Dependencies (arch only for now)
+if [[ "${BASE_DISTRO:-}" == "arch" ]]; then
```
0xSolanaceae, line 41: the Qt6 deps are only installed on Arch. On Fedora or Debian this block is skipped entirely, but line 77 still writes `Current=caelestia` and line 119 still prints "SDDM theme installed", with nothing having checked that the theme can actually run. should this fail early on non-Arch, or handle the other distros like the other step scripts do?

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,119 @@
...
+ok "SDDM config drop-in created."
+
+# Posthook registration
+POSTHOOK_CMD="sudo $SYNC_SCRIPT --posthook"
+CLI_JSON="$HOME/.config/caelestia/cli.json"
+
+if command -v python3 &>/dev/null; then
+    python3 - "$CLI_JSON" "$POSTHOOK_CMD" <<'PYEOF'
+import json, sys, os
+cli_path, hook_cmd = sys.argv[1], sys.argv[2]
```
0xSolanaceae, line 139: this whole section assigns postHook unconditionally for the wallpaper and theme sections, so any hook the user already had is silently replaced. Worth preserving or merging it. Also, when python3 is missing the whole block is skipped, but line 119 still reports success.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,119 @@
...
+    config[section]["postHook"] = hook_cmd
+os.makedirs(os.path.dirname(cli_path), exist_ok=True)
+with open(cli_path, "w") as f:
+    json.dump(config, f, indent=4)
+PYEOF
+    ok "Posthook registered in cli.json"
+fi
+
+# Passwordless sudo for sync
+SUDOERS_FILE="/etc/sudoers.d/caelestia-sddm-sync"
```
0xSolanaceae, line 164: this creates a permanent NOPASSWD root grant at install time. The target path is root owned so the blast radius is limited, but nothing removes it later, since uninstall.sh is untouched here. Could uninstall remove this?

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,119 @@
...
+fi
+
+# Initial sync
+if caelestia_sudo "$SYNC_SCRIPT"; then
+    ok "Initial sync complete."
+else
+    warn "Initial sync had warnings (non-fatal)."
+fi
+
+ok "SDDM theme installed."
```
0xSolanaceae: this prints even when posthook reg was skipped for missing python3, please gate this success message to see if it actually did install it

src/sddm/themes/mini/assets/google-sans-flex/GoogleSansFlex.ttf
0xSolanaceae, line 1: this is duplicated again

src/sddm/sync.sh
```diff
@@ -0,0 +1,206 @@
...
+    fi
+
+    if ! sudo -H -u "$REAL_USER" head -c "$((max_bytes + 1))" -- "$src" | cat > "$read_tmp"; then
+        fail_cleanup
+        return
+    fi
+
+    actual_size="$(stat -c '%s' "$read_tmp")"
+
+    if [ "$actual_size" -gt "$max_bytes" ]; then
```
0xSolanaceae: the code standard in repo is to use double square brackets for code security reasons, please convert [ to [[

src/sddm/sync.sh
```diff
@@ -0,0 +1,206 @@
...
+if [ -z "$REAL_HOME" ] || [ "$REAL_HOME" = "/" ]; then
+    echo "ERROR: Could not determine a valid home directory for $REAL_USER." >&2
+    exit 1
+fi
+
+CAEL_STATE="$REAL_HOME/.local/state/caelestia"
+THEME_DIR="/usr/share/sddm/themes/caelestia"
+
+# Clear SDDM greeter QML cache
+rm -rf /var/lib/sddm/.cache/sddm-greeter-qt6
```
0xSolanaceae: This runs on every posthook, so every wallpaper or theme change wipes `/var/lib/sddm/.cache/sddm-greeter-qt6`. that doesn't seem to be intentional

src/sddm/sync.sh
```diff
@@ -0,0 +1,206 @@
...
+# 4. Sync Wallpaper LAST
+WALLPAPER_SRC="$CAEL_STATE/wallpaper/current"
+MAX_WALLPAPER_BYTES=$((50 * 1024 * 1024))
+MAX_VIDEO_WALLPAPER_BYTES=$((250 * 1024 * 1024))
+WALLPAPER_DEST="$THEME_DIR/assets/background"
+
+# Video wallpapers are synced with a proper extension (needed for SDDM to recognize it)
+if command -v file >/dev/null 2>&1; then
+    WALLPAPER_MIME="$(sudo -H -u "$REAL_USER" file -b --mime-type -L "$WALLPAPER_SRC" 2>/dev/null || true)"
+    case "$WALLPAPER_MIME" in
```
0xSolanaceae, line 157: `$WALLPAPER_MIME` is switched on, then switched on again inside itself, then a third switch on the derived extension. could this collapse into one lookup?

src/sddm/themes/full/components/shapes/material-shapes.js
0xSolanaceae, line 1: all 15 files here are duplicates of `themes/full/components/shapes`. please pull from the already instantiated copies

src/sddm/themes/mini/metadata.desktop
0xSolanaceae, line 1: this variant is missing ConfigFile=theme.conf, and uses X-SDDM-Theme-InterfaceVersion=2 where mini uses Theme-API=2.0. The one we actually install is the inconsistent one. Which is correct for Qt6 SDDM?

src/sddm/themes/full/theme.conf
```diff
@@ -0,0 +1,69 @@
...
+# ============================================================================
+# WELCOME MESSAGE
+# ============================================================================
+# Enable welcome message [true | false]
+enableWelcomeMessage=false
+
+# Welcome message background opacity [0.0 to 1.0]
+welcomeBgBlurAmount=0.7
+welcomeColorOpacity=0.7
+# mainCard background opacity and blur [0.0 to 2.0]
```
0xSolanaceae: this is a pre-filled copy of theme.conf.template in the same folder, kept as the fallback before the first sync overwrites it. Worth a one-line comment saying so, and the two have already drifted here: the template says [0.0 to 1.0], this says [0.0 to 2.0].

0xSolanaceae (changes requested):
thanks for contributing! couple of notes: GoogleSansFlex.ttf is committed twice, please use the font ttf already available in the repo. also, uninstall.sh is untouched, so the sudoers file, and the cli.json postHooks are not removed on uninstall. please fix this

aroaxinping:
@0xSolanaceae addressed all review feedback in 4c8c591f:

Duplicate fonts removed from both theme variants, copied from the existing lockscreen source at install time. Duplicate shapes removed from mini, same approach. Added SDDM_THEME_VARIANT selector (full/mini) to the installer menu. Multi-distro deps handled for Fedora (dnf) and Debian (apt) alongside Arch. Posthooks merged instead of overwritten. Missing python3 now warns instead of silently skipping. Success message gated on actual installation status. sync.sh converted to double brackets, cache wipe removed, MIME switch collapsed. metadata.desktop aligned between variants. theme.conf comment drift fixed. uninstall.sh now cleans up the sudoers drop-in, SDDM theme files, config drop-in, cli.json posthooks, and template config.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,173 @@
...
+if os.path.exists(cli_path):
+    with open(cli_path) as f:
+        config = json.load(f)
+for section in ("wallpaper", "theme"):
+    if section not in config:
+        config[section] = {}
+    existing = config[section].get("postHook", "")
+    if existing and hook_cmd not in existing:
+        config[section]["postHook"] = existing + " && " + hook_cmd
+    else:
```
0xSolanaceae, line 152: This else catches both "no existing hook" and "hook already present", and assigns hook_cmd alone in the second case, so a pre-existing hook is dropped on re-install. I ran this block twice against {"postHook": "notify-send hi"}:
`after 1st run:  "notify-send hi && sudo /usr/share/sddm/.../sync.sh --posthook"
after 2nd run:  "sudo /usr/share/sddm/.../sync.sh --posthook"`
The intent was idempotency, so this branch should leave existing unchanged rather than replace it. The update path re-runs steps, so this fires on a normal re-install.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,173 @@
...
+if [[ -d "$INSTALL_DIR" ]]; then
+    caelestia_sudo rm -rf "$INSTALL_DIR"
+fi
+
+caelestia_sudo mkdir -p "$INSTALL_DIR/scripts"
+caelestia_sudo cp -r "$THEME_SOURCE"/* "$INSTALL_DIR/"
+caelestia_sudo cp "$SRC_DIR/sync.sh" "$INSTALL_DIR/scripts/"
+
+caelestia_sudo mkdir -p "$INSTALL_DIR/assets/google-sans-flex"
+if [[ -f "$FONT_SOURCE" ]]; then
```
0xSolanaceae, line 94: No else here, so if the font source is missing the theme installs without a font and nothing is reported. Since this path points into another package's directory, a warn would make that failure visible.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,173 @@
...
+    caelestia_sudo cp "$FONT_SOURCE" "$INSTALL_DIR/assets/google-sans-flex/GoogleSansFlex.ttf"
+fi
+
+if [[ "$VARIANT" == "mini" && -d "$SRC_DIR/themes/full/components/shapes" ]]; then
```
0xSolanaceae: Same shape: a missing source silently installs a mini theme with no shapes, which will not render. Also worth a comment noting that mini now depends on the full variant's directory, so the two are coupled and a future restructure of full/ would break mini silently.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,173 @@
...
+GreeterEnvironment=QML_XHR_ALLOW_FILE_READ=1
+
+[Theme]
+Current=caelestia
+DROPIN
+ok "SDDM config drop-in created."
+
+POSTHOOK_CMD="sudo $SYNC_SCRIPT --posthook"
+CLI_JSON="$HOME/.config/caelestia/cli.json"
+POSTHOOK_OK=false
```
0xSolanaceae: This is set at line 149 but never read. Looks like it was the original gate before ALL_OK took over, so it can go.

scripts/05-sddm-theme.sh
```diff
@@ -0,0 +1,173 @@
...
+fi
+
+if caelestia_sudo "$SYNC_SCRIPT"; then
+    ok "Initial sync complete."
+else
+    warn "Initial sync had warnings (non-fatal)."
+    ALL_OK=false
+fi
+
+if [[ "$ALL_OK" == "true" ]]; then
```
0xSolanaceae, line 181: ALL_OK=false only changes the closing message, the script still exits 0, so the TUI records this step as OK for a run that installed with warnings. Worth deciding if that should surface as a WARN status instead.

0xSolanaceae (changes requested):
few more, thanks!

aroaxinping:
@0xSolanaceae addressed the second round in 708b094f:

Exit 1 when ALL_OK is false so the TUI records the step as failed instead of silently passing. Removed the unused POSTHOOK_OK variable. Font and shape source missing now warns instead of installing a broken theme silently. Added comment noting mini depends on full's shape components. Fixed posthook idempotency so re-install preserves pre-existing hooks when the SDDM hook is already registered.

0xSolanaceae (approved):
I pushed 693edd1d to your branch. It drops the exit 1 and keeps the step exiting 0, because the TUI already reads the [WARN] markers in a step's output and reports WARN. The non-zero exit was turning a correct WARN into FAILED and opening the Retry dialog, where Retry just re-runs a step that already installed

thanks for your work!

PR #707, opened 2026-09-11
feat: cava prebuild integration

ladybug-me:

> [!WARNING]

## Type

- [ ] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

### DIFF
--- a/installer/distro/arch/packages.sh
+++ b/installer/distro/arch/packages.sh
@@ -90,11 +90,13 @@ if [[ "$PACKAGE_GROUP" == "all" || "$PACKAGE_GROUP" == "themes" ]]; then
     fi
 fi
 
-# libcava and darkly are installed from the AUR. Darkly is published as the
-# prebuilt darkly-bin package, so it is never compiled on the machine; libcava
-# has no -bin package and compiles from source.
+# Developer SDK (libcava) extracted from prebuilt release assets.
 if [[ "$PACKAGE_GROUP" == "all" || "$PACKAGE_GROUP" == "core" ]]; then
-    PACKAGES+=(libcava)
+    if curl -fsSL "https://github.com/ladybug-me/cava/releases/download/continuous/cava-x86_64-arch.tar.gz" | sudo tar -C /usr -xzf - --exclude='bin' 2>/dev/null; then
+        log "Installed prebuilt CAVA SDK from release."
+    else
+        PACKAGES+=(libcava)
+    fi
 fi
 if [[ "$PACKAGE_GROUP" == "all" || "$PACKAGE_GROUP" == "themes" ]]; then
     if [[ "$INSTALL_DARKLY" == "true" ]]; then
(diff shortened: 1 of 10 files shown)


installer/distro/arch/packages.sh
```diff
@@ -90,11 +90,13 @@ if [[ "$PACKAGE_GROUP" == "all" || "$PACKAGE_GROUP" == "themes" ]]; then
     fi
 fi
 
-# libcava and darkly are installed from the AUR. Darkly is published as the
-# prebuilt darkly-bin package, so it is never compiled on the machine; libcava
-# has no -bin package and compiles from source.
+# Developer SDK (libcava) extracted from prebuilt release assets.
 if [[ "$PACKAGE_GROUP" == "all" || "$PACKAGE_GROUP" == "core" ]]; then
-    PACKAGES+=(libcava)
+    if curl -fsSL "https://github.com/ladybug-me/cava/releases/download/continuous/cava-x86_64-arch.tar.gz" | sudo tar -C /usr -xzf - --exclude='bin' 2>/dev/null; then
```
0xSolanaceae: This same fetch-and-extract line is copied four times:

> installer/distro/arch/packages.sh:95
> installer/distro/debian/packages.sh:181
> installer/distro/fedora/packages.sh:192
> .github/workflows/version-release.yml:222

Only the asset name changes. They've already drifted too: the installer copies pass --exclude='bin', the workflow one doesn't. put it  into one helper that takes the distro name, e.g. install_cava_sdk arch, and calls it from all four

shell/plugin/src/Caelestia/Services/cavaprovider.cpp
```diff
@@ -108,7 +108,11 @@ void CavaProcessor::initCava() {
         return;
     }
 
+#ifdef CAVA_SCALING_LINEAR
+    m_plan = cava_init(m_bars, ac::SAMPLE_RATE, 1, 1, 0.85, 50, 10000, CAVA_SCALING_LINEAR);
```
0xSolanaceae: The same cava_init argument list is written twice, differing only in the last argument. If the two signatures force two calls, that's fine, but the repeated 0.85, 50 and 10000 can drift apart. Give them names and use them in both calls.

0xSolanaceae (changes requested):
2 fixes, thanks!

installer/distro/arch/packages.sh
ladybug-me: Done

shell/plugin/src/Caelestia/Services/cavaprovider.cpp
ladybug-me: Done

ladybug-me:
All done @0xSolanaceae

0xSolanaceae (approved):
LGTM!

PR #768, opened 2026-09-17
i18n(ru): review the Russian catalogue — 1322 strings

plagness:
## What does this change?

A full review pass over the Russian catalogue: **1322 of 2148** messages rewritten (1320 in the first commit, 3 tab labels in the second). The catalogue came in from Crowdin largely machine-translated, and several classes of error made the shell read wrong to a native speaker or broke panel layout.

Since most reviewers here will not read Russian, here is a sample with literal back-translations of what the old text actually said:

| Source | Was | …which reads as | Now |
| --- | --- | --- | --- |
| `Shell` | Ракушка | "a seashell" | Оболочка |
| `Clam Shell` | Поджигательная раковина | "arson seashell" | Ракушка |
| `Background` | Справочная информация | "background as in a person's biography" | Фон |
| `Volume` | Объем | "volume as in cubic capacity" | Громкость |
| `Development` | Развитие | "personal growth" | Разработка |
| `drivers` | водители | "people who drive cars" | драйверы |
| `Toasts` | Токсы | not a word | Тосты |
| `Keep Awake` | Продолжить пробуждение | "resume awakening" | Не засыпать |
| `Cookie 4-Sided` | Печенье 4-х печенье | "cookie 4-of cookie" | Печенье, 4 грани |
| `All up to date!` | Всё в курсе! | "everyone is informed!" | Всё обновлено! |
| `Ollama` | Оллама | transliterated product name | Ollama |
| `Quickshell` | Быстрая оболочка | "fast shell" | Quickshell |
| `Spectacle` | Спектакль | "a theatre show" | Spectacle |
| `OpenRouter` | Открытый маршрутизатор | "open network router" | OpenRouter |

Other proper nouns that were being translated and are now left alone: `Claude`, `Claude Code`, `Gemini`, `Wallhaven`, `BTree`, `EasyEffects`, `Bad Apple`, `SF Mono`, `CaskaydiaCove NF`, and the conventional-commit types (`feat`, `fix`, `perf`, `chore`, …).

English contractions were split mid-word, leaving a stray apostrophe inside the Russian text — `Couldn't save API key` → "Не удалось**'т** сохранить ключ API", `doesn't broadcast` → "не**'т** транслирует", `Google's` → "Google**'с**", `it's in range` → "она**'с** в диапазоне". 40+ strings.

One string also stated the opposite of the truth: *Untranslated text falls back to English* was translated as "falls back to **Russian**".

## Screenshots (if UI change)

Text-only change, so the table above is the substantive diff. One layout fix did come from looking at it on screen, though: in the sidebar the `Notifications` and `AI Assistant` tabs ran into each other with no gap, and the dashboard's `Performance` tab was by far the widest of its five. Those three are now "Ассистент" and "Нагрузка" (the ordinary Russian word for CPU/GPU/memory load, 8 characters against 18); the full "ИИ-ассистент" and "Производительность" stay on the Settings pages, where there is room.

## How did you test it?

Running live on KDE Plasma 6.7 / KWin, Caelestia built from source on NixOS, 2560x1440 at 1.25 scale — which is how the tab overflow above was noticed.

Layout was measured, not eyeballed. Russian runs longer than English, and many of these strings sit in fixed-width places — bar entries, switch rows, popout headers. Short labels had grown well past their slots: `Fit` → "По размеру", `Other` → "Прочие вопросы", `General` → "Общие положения", `Used` → "Использовано".

- RU/EN character ratio over the whole catalogue: **1.156 → 1.076**
- short labels (≤20 chars) overshooting their source by more than 50%: **61 → 1**
- worst-case growth for a short label: **+25 → +12** characters

Checked with a script over the catalogue, plus reading every changed string:

- no unfinished entries (**was 25**)
- placeholder sets match between source and translation for every message except the three noted below
- XML parses; `lrelease` reports **2148 finished / 0 unfinished**
- numerus forms untouched (0 lines changed)
- `caelestia_ru.qm` recompiled with `lrelease` from Qt 6.11.2

## Type

- [ ] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [x] Other — translations

## Notes for reviewers

**Three strings deliberately drop a placeholder.** The English builds plurals by appending `s` through a second placeholder — `%1 device%2 available`, `%1 notification%2`, `notification%1`. That cannot work in Russian, which has three plural forms (1 устройство / 2 устройства / 5 устройств), so no single string with an appended suffix is correct. They now read as *"Доступно устройств: %1"* ("devices available: %1"), which is correct for every count. The real fix belongs upstream in the QML, by switching those call sites to numerus forms — I left that alone rather than guessing at your preference.

**Crowdin.** Since the catalogue is synced with Crowdin, this may need to land there rather than here. Happy to move it if that is the preferred path — I went with a PR because the guide in `shell/translations/README.md` documents the in-repo route as well.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


0xSolanaceae:
excellent, thank you for your contribution for us non russian speakers!

PR #770, opened 2026-09-17
fix: set asynchronous to false

ladybug-me:
So the thing I found is, KDE's underlying icon loading engine (`KIconLoader`) is not thread-safe. When Qt Quick uses `asynchronous: true` on an `IconImage`, it spawns multiple background threads to load icons concurrently (e.g., when the Window Switcher or Overview opens). This concurrent access to the non-thread-safe engine caused the `SIGSEGV` crash. (found through core dump)

Now, a decision has to be made.
`asynchronous: false`, for obvious reasons, will cause some stutter on slow machines.

The alternative and the already in place fix / workaround is the automatic restart of shell anyways...


shell/components/effects/ColouredIcon.qml
```diff
@@ -10,7 +10,7 @@ IconImage {
 
     required property color colour
 
-    asynchronous: true
+    asynchronous: false // FIX 759: unsafe with KIconLoader
```
0xSolanaceae, line 13: wasn't this already fixed by PR #762 ? if this is a different crash, open a new issue to refer to or rm issue. probably rm issue tag either way tbh, not necessary

shell/components/images/CachingIconImage.qml
```diff
@@ -38,7 +38,7 @@ Item {
 
         IconImage {
             source: root.source
-            asynchronous: true
+            asynchronous: false // FIX 759: unsafe with KIconLoader
```
0xSolanaceae, line 41: The Loader at line 22 still has asynchronous: true, so the path for this component does not change. is that on purpose? also putting "icon lookups decode on the GUI thread" here (or in one shared wrapper) keeps the invariant in one place instead of 15 files, which is what the current version does

shell/modules/overview/WindowGrid.qml
```diff
@@ -652,7 +652,7 @@ Item {
 
                                     IconImage {
                                         implicitSize: Math.round(titleText.implicitHeight * 1.1)
-                                        asynchronous: true
+                                        asynchronous: false // FIX 759: unsafe with KIconLoader
```
0xSolanaceae, line 655: This one usually is not a KIconLoader lookup at all: windowIconSource() returns "file://" + wp first (line 104-111), so for any window with an extracted _NET_WM_ICON this now sync-decodes a plain PNG on the GUI thread with no KIconLoader benefit. Same applies to the dock's WinIcons.sourceFor. Gating on the source being an image://icon/ lookup would fix the crash surface and leave the cheap async paths alone.

shell/modules/launcher/items/AppTile.qml
```diff
@@ -68,7 +68,7 @@ Item {
             id: icon
 
             anchors.horizontalCenter: parent.horizontalCenter
-            asynchronous: true
+            asynchronous: false // FIX 759: unsafe with KIconLoader
```
0xSolanaceae, line 71: this is the launcher grid, so the stutter & slowness happens on every open. also please add in a config toggle for this so users can choose to switch between the two

0xSolanaceae (changes requested):
good, just fix references and add a config toggle in settings if it has the potential to cause performance drops

ladybug-me:
Thx for the review but closing this one in favor of 406176e.

PR #784, opened 2026-09-20
Don't launch Spotify as a side effect of spicetify theming

prajwalkarna:
spicetify apply patches Spotify's app files directly on disk, which causes Spotify to launch if it isn't already running. Since caelestia-color runs this on every wallpaper/scheme change (not just at login), this meant Spotify would pop open unexpectedly any time the color scheme regenerated, even if the user never opened it.

Gate the spicetify apply call on Spotify already being running via pgrep, so theming still happens correctly whenever Spotify is open, but no longer launches it as a side effect.

## What does this change?

`spicetify apply` patches Spotify's files on disk, which causes
Spotify to launch even if it wasn't running. Since caelestia-color
calls this on every wallpaper/scheme change (not just at login),
changing your wallpaper with Spotify closed would unexpectedly
launch it. This adds a check so `spicetify apply` only runs when
Spotify is already open.

## How did you test it?

Closed Spotify, then triggered a scheme change via a wallpaper
change - confirmed Spotify no longer launches. Then opened Spotify
manually, triggered another scheme change, and confirmed it still
themes correctly (colors update, no launch since it was already
running).

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Config / theme
- [ ] Docs
- [ ] Refactor
- [ ] Other

## Notes for reviewers

None - single-line change, gated on `pgrep -x spotify` so theming
behavior is otherwise identical.


0xSolanaceae:
@prajwalkarna good catch, thanks for the PR!

